Issue #13, opened 2015-03-10
Remove Client UI


jazeren1:
The Client UI has since been broken and needs to be redesigned.  It should be pulled out of the codebase until it can be properly redeveloped.

PR #16, opened 2015-03-11
#13 Removing broken client UI

jazeren1:
@sagarsane Removing the client UI.  From what I could tell we just needed to remove the jcr_root directory.  Let me know if I missed something OR if there is something in the jcr_root folder we need to keep.


sagarsane:
Thanks @jazeren1. Should we also update the `filter.xml` ?

jazeren1:
@sagarsane do we even need a filter xml at all?  I removed both paths that are mentioned here so maybe we can just remove the filter xml for now.

```
<workspaceFilter vesion="1.0">
    <filter root="/apps/grabbit" />
    <filter root="/etc/clientlibs/grabbit" />
</workspaceFilter>
```

sagarsane:
I think we will need to keep `/apps/grabbit` as the install path is `/apps/grabbit/install` ..

jazeren1:
@sagarsane I updated the filter.xml to remove the /etc/ entry.

sagarsane:
Testing looks good :+1:

Issue #23, opened 2015-04-17
Add "excludePaths" per path


jdigger:
Something like
```json
{
    "path" : "/content/someContent",
    "excludePaths" :
        [
            "subPath1",
            "subPath2/xyz"
        ],
}
```
To keep things clear/simple, the `excludePaths` should only allow relative paths to the `path` for that block. For example the following would be illegal:
```json
{
    "path" : "/content/someContent",
    "excludePaths" :
        [
            "/content/otherContent",
            "/content/someContent/apath"
        ],
}
```
An open question: Should "." be legal?

sagarsane:
I guess `.` would effectively be a `no-op` right (especially if we only allow sub-paths)? If so, I think we could go eitherway. Legal or Illegal.

jdigger:
To work with specifics; given the following path structure:
```
/a
/a/b
/a/c
```
How would you express the following?

1. Get everything in `/a`
1. Get all children under `/a`, but not `/a` itself
1. Get only `/a` itself, not its children

I expect it would like

1. `{ "path": "/a" }`
1.  `{ "path": "/a", "excludePaths": ["."] }`
1.  `{ "path": "/a/." }` or `{ "path": "/a", "excludePaths: ["*"] }`

PR #29, opened 2015-05-01
Exclude paths

masroormohammed:
Issue #23 Add Exclude Paths 

- [x] Coding Completed
- [x] Manual Testing Completed
- [x] Jackalope Test Scripts

@jdigger @sagarsane


sagarsane:
@jazeren1

sagarsane:
@masroormohammed take a look at how to change spacing in your IDE from "tabs" to "spaces". http://stackoverflow.com/a/408449

The code convention here is 4spaces instead of tabs.

masroormohammed:
@sagarsane fixed indentation issues

masroormohammed:
@jbyrnessnip @sikarwarvimal

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -137,11 +138,27 @@ class GrabbitConfiguration {
...
     static class PathConfiguration {
+        private static final String COMMA = ","
+        private static final String SLASH = "/"
+        private static final String EMPTY = ""
         String path
+        Collection<String> excludePaths
         Collection<String> workflowConfigIds
+        
+        Collection<String> modifyExcludePaths(@Nonnull Collection<String> excludePaths)
+        {
```
jdigger: Minor: Doesn't follow the style conventions in the rest of the code base. If you're using IDEA and generate the project with `./gradlew idea` then it should be automatically set up to format things consistently. (Including things like tabs vs spaces.)

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -137,11 +138,27 @@ class GrabbitConfiguration {
...
+            Collection<String> temp = excludePaths
+            String modifiedExcludePaths = EMPTY;
+            for(String append : temp)
+            {
+                modifiedExcludePaths += this.path+SLASH+append+COMMA;
+            }
+            return modifiedExcludePaths.split(COMMA) as Collection<String>
```
jdigger: This entire method can be simply written:
```groovy
return excludePaths.collect { "${path}/${it}" }
```

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -137,11 +138,27 @@ class GrabbitConfiguration {
...
     @CompileStatic
     static class PathConfiguration {
+        private static final String COMMA = ","
+        private static final String SLASH = "/"
+        private static final String EMPTY = ""
         String path
+        Collection<String> excludePaths
         Collection<String> workflowConfigIds
+        
+        Collection<String> modifyExcludePaths(@Nonnull Collection<String> excludePaths)
```
jdigger: An instance method called "modify*" implies that it will modify the instance, which this doesn't do. It should be made `static` and renamed to say what it actually does (`toAbsolutePaths` or something).

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
jdigger: One of the reasons I want those unit tests (which in this case wouldn't need JCR mocking), is what would/should happen if this received input like `[null, "foo", "/bar", "./bunny", "\\yikes", "co,ma"]`? (There's nothing preventing a comma being a part of a node's name)

jdigger: To be very specific about watching out for assumptions about the contents of an exclude path, see http://www.ietf.org/rfc/rfc2396.txt which talks about what's legal in a URI path. (Specifically section 2.2)

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
```diff
@@ -179,6 +180,7 @@ class ClientBatchJob {
                     "${PORT}"             : serverBuilder.port,
                     "${USERNAME}"         : credentialsBuilder.username,
                     "${PASSWORD}"         : credentialsBuilder.password,
+                    "${EXCLUDE_PATHS}"	  : pathConfiguration.excludePaths.join("|"),
```
jdigger: Is this safe, since technically `"|"` can be in a path?

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJobExecutionListener.groovy
```diff
@@ -120,7 +121,8 @@ class ClientBatchJobExecutionListener implements JobExecutionListener{
 
         final String encodedContentAfterDate = URLEncoder.encode(contentAfterDate, 'utf-8')
         final String encodedPath = URLEncoder.encode(path, 'utf-8')
-        final grabPath = "http://${host}:${port}/grabbit/job?path=${encodedPath}&after=${encodedContentAfterDate}"
+        final String encodedExcludedPaths = URLEncoder.encode(excludePaths, 'utf-8')
+        final grabPath = "http://${host}:${port}/grabbit/job?path=${encodedPath}&excludePaths=${encodedExcludedPaths}&after=${encodedContentAfterDate}"
```
jdigger: May want to start looking at using something like `org.apache.http.client.utils.URIBuilder` so that edge-cases around URL encoding are taken care of (like, what if `port` is blank because it's been mapped to 80?).

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/ServerBatchJob.groovy
```diff
@@ -85,14 +86,19 @@ class ServerBatchJob {
...
 
         PathBuilder(ConfigurationBuilder configurationBuilder) {
             this.configurationBuilder = configurationBuilder
         }
 
-        ContentAfterDateBuilder andPath(String path) {
+        ContentAfterDateBuilder andPath(String path, String excludePaths) {
             if(path == null) throw new IllegalArgumentException("path == null")
             this.path = path
+            if(!excludePaths.isEmpty() && excludePaths != null)
```
jdigger: Logic is backwards. If `excludePaths == null` then this will give an NPE.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
```diff
@@ -32,18 +31,54 @@ import javax.jcr.Node as JcrNode
...
-    @Override
+	
+	private String excludePaths
+	private String[] excludePathList
+	private static final String[] EMPTY_ARRAY = new String[0];
+	
+	public void setExcludePaths(String excludePaths) 
+	{
+        this.excludePaths = excludePaths
+        if(!this.excludePaths.isEmpty() && this.excludePaths != null)
```
jdigger: What happens if `excludePaths == null`?

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
```diff
@@ -32,18 +31,54 @@ import javax.jcr.Node as JcrNode
...
+	@Override
     JcrNode read() throws Exception, UnexpectedInputException, ParseException, NonTransientResourceException {
         Iterator<JcrNode> nodeIterator = theNodeIterator()
         if(nodeIterator == null) throw new IllegalStateException("nodeIterator must be set.")
-
-        if(nodeIterator.hasNext()) {
-            nodeIterator.next()
+        JcrNode returnNode = null;
+        while(true)
+        {
```
jdigger: Minor, but need to fix formatting to be consistent with the rest of the project.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
```diff
@@ -32,18 +31,54 @@ import javax.jcr.Node as JcrNode
...
+                JcrNode node = nodeIterator.next()
+                if(!isPathInExcludedList(node.getPath()))
+                {
+                    returnNode = node;
+                    break;
+                }
+            }
+            else
+            {
+                break;
```
jdigger: All these `breaks` are making it too hard to follow. Simplify the control logic.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
```diff
@@ -32,18 +31,54 @@ import javax.jcr.Node as JcrNode
...
+	private String excludePaths
+	private String[] excludePathList
+	private static final String[] EMPTY_ARRAY = new String[0];
+	
+	public void setExcludePaths(String excludePaths) 
+	{
+        this.excludePaths = excludePaths
+        if(!this.excludePaths.isEmpty() && this.excludePaths != null)
+            this.excludePathList = excludePaths.split("\\|");
+		else
```
jdigger: Why is this indented so far in?

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
jdigger: Why not simply return the node when it's found?

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
```diff
@@ -32,18 +31,54 @@ import javax.jcr.Node as JcrNode
...
+            {
+                break;
+            }
         }
-        else {
-            null
+        returnNode
+    }
+    
+    private boolean isPathInExcludedList(String path)
```
jdigger: Can be rewritten as
```groovy
return excludePathList.contains(path)
```

grabbit/src/main/groovy/com/twcable/grabbit/server/services/impl/DefaultServerService.groovy
```diff
@@ -50,7 +50,7 @@ class DefaultServerService implements ServerService{
     ConfigurableApplicationContext configurableApplicationContext
 
 
-    void getContentForRootPath(@Nonnull String path, @Nullable String afterDateString, @Nonnull ServletOutputStream servletOutputStream) {
+    void getContentForRootPath(@Nonnull String path, @Nonnull String excludePaths, @Nullable String afterDateString, @Nonnull ServletOutputStream servletOutputStream) {
```
jdigger: Are you verifying the new precondition?

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJobExecutionListener.groovy
jdigger: FWIW, on the specific example of what if `port` is blank, maybe a better solution to that is to explicitly map a blank port value to `80`.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/ServerBatchJob.groovy
```diff
@@ -85,14 +86,19 @@ class ServerBatchJob {
...
         PathBuilder(ConfigurationBuilder configurationBuilder) {
             this.configurationBuilder = configurationBuilder
         }
 
-        ContentAfterDateBuilder andPath(String path) {
+        ContentAfterDateBuilder andPath(String path, String excludePaths) {
             if(path == null) throw new IllegalArgumentException("path == null")
             this.path = path
+            if(excludePaths != null && !excludePaths.isEmpty())
+            {
```
jdigger: Trivial: to be consistent, the `{` should be on the line with the `if`

jdigger:
`gw clean check` gives
```shell
com.twcable.grabbit.client.batch.workflows.WorkflowManagerSpec > Should manage multithreading correctly FAILED
    org.spockframework.runtime.SpockComparisonFailure at WorkflowManagerSpec.groovy:59
```

jdigger:
Never mind... It looks transient (which makes me a little worried for multithreading code).

sagarsane:
I have seen this happen when executing multiple gradle tasks at once (i.e., `gradlew build` or `gradlew tR` seems to not fail). May be the test can be updated to be more robust.

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -25,6 +25,93 @@ import spock.lang.Unroll
 class GrabbitConfigurationSpec extends Specification {
 
     @Unroll
+    def "Should return exclude paths"() {
+        when:
+        def output = GrabbitConfiguration.create(input)
+
+        then:
+        output instanceof GrabbitConfiguration
```
jdigger: I don't understand what this is testing. What other possible state could this be in?

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -25,6 +25,93 @@ import spock.lang.Unroll
...
+            {
+                "serverUsername" : "admin",
+                "serverPassword" : "admin",
+                "serverHost" : "localhost",
+                "serverPort" : "4503",
+                "deltaContent" : false,
+                "pathConfigurations" :  [
+                    {
+                        "path" : "/content/a/b",
+						"excludePaths" : ["/a", "\\b", "./c"],
```
jdigger: What's going on with the indenting here? Do you still have hard tabs?

jdigger: #23 says that absolute paths are explicitly illegal for exclusion. Why doesn't this blow up?

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -77,6 +78,20 @@ class GrabbitConfiguration {
                 pathConfigurations
         )
     }
+	
+	private static Collection<String> nonEmptyCollection(Collection<String> excludePaths,ConfigurationException.Builder errorBuilder) {
+		if(excludePaths == null || excludePaths.isEmpty())
+			return excludePaths
```
jazeren1: @masroormohammed this return can either return null or an empty collection.  I think you might want to be specific about what you want to return here in the case that there are no exclude paths.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
```diff
@@ -179,6 +180,7 @@ class ClientBatchJob {
                     "${PORT}"             : serverBuilder.port,
                     "${USERNAME}"         : credentialsBuilder.username,
                     "${PASSWORD}"         : credentialsBuilder.password,
+                    "${EXCLUDE_PATHS}"    : pathConfiguration.excludePaths.join("<sp>"),
```
jazeren1: @masroormohammed I'm not familiar with this notation "<sp>".  Does this represent a space?

masroormohammed:
@jdigger ready for another round of review

README.md
```diff
@@ -72,10 +72,16 @@ A `json` configuration file of following format is used to configure Grabbit.
...
+            "path" : "/content/someContent",
+            "excludePaths" : "/content/someContent/someExcludeContent"
         },
         {
             "path" : "/content/dam/someDamContent",
+            "excludePaths":
+                [
+                    "/content/someContent/someExcludeContent",
+                    "/content/someContent/someOtherExcludeContent"
+                ],
```
jdigger: This is still showing absolute paths in `"excludePaths"`.

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -25,6 +25,125 @@ import spock.lang.Unroll
...
+        when:
+        def expectedOutput = ["/content/a/b/d", "/content/a/b/e/ab", "/content/a/b/q/fg"] as Collection<String>
+        def actualOutput = GrabbitConfiguration.create(input)
+
+        then:
+        actualOutput instanceof GrabbitConfiguration
+
+        actualOutput.pathConfigurations.excludePaths.get(0) == expectedOutput
+
+        where:
```
jdigger: What value is the `where:` block providing, since there's only one datapoint?

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -25,6 +25,125 @@ import spock.lang.Unroll
 class GrabbitConfigurationSpec extends Specification {
 
     @Unroll
```
jdigger: What value is `@Unroll` providing since there's no variable substitution happening in the feature name?

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -25,6 +25,125 @@ import spock.lang.Unroll
...
+    def "Should return exclude paths"() {
+        when:
+        def expectedOutput = ["/content/a/b/d", "/content/a/b/e/ab", "/content/a/b/q/fg"] as Collection<String>
+        def actualOutput = GrabbitConfiguration.create(input)
+
+        then:
+        actualOutput instanceof GrabbitConfiguration
+
+        actualOutput.pathConfigurations.excludePaths.get(0) == expectedOutput
```
jdigger: Stylistic: Instead of `.get(0)`, use `.first()` -- it does effectively the same thing, but feels less like you're using a magic number.

jdigger: Logically what you're really doing is selecting the first value from `pathConfigurations` and then getting its `excludePaths` value. So it *should* read
```groovy
actualOutput.pathConfigurations.first().excludePaths == expectedOutput
```
The reason the current way works is because since you're not selecting a specific member of the `pathConfigurations` collection, it's doing a "spread" and selecting all of its elements that contain `excludePaths`. It's pretty weird behavior coming from a static background, but common in dynamic contexts, such as XPath. Since here's you're using Groovy in its dynamic form, it's doing the spread. However it would be much more clear if you expressed your intent better by selecting the first `pathConfigurations` instead of relying on the result of it doing the spread combinations.

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -25,6 +25,125 @@ import spock.lang.Unroll
 class GrabbitConfigurationSpec extends Specification {
 
     @Unroll
+    def "Should return exclude paths"() {
+        when:
+        def expectedOutput = ["/content/a/b/d", "/content/a/b/e/ab", "/content/a/b/q/fg"] as Collection<String>
```
jdigger: Why do you have `as Collection<String>`?

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -25,6 +25,125 @@ import spock.lang.Unroll
 class GrabbitConfigurationSpec extends Specification {
 
     @Unroll
+    def "Should return exclude paths"() {
```
jdigger: That's not a very descriptive description of the feature being verified...

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -25,6 +25,125 @@ import spock.lang.Unroll
...
+                    {
+                        "path" : "/content/a/b",
+                        "excludePaths" : ["d", "e/ab", "q/fg"],
+                        "workflowConfigIds" : []
+                    }
+                ]
+			}
+			"""
+        ]
+    }
```
jdigger: Bringing the previous comments together, here's a potential way of rewriting the feature spec:
```groovy
def "exclude paths should be resolved to their parent path"() {
    given:
    def input =
            """
        {
            "serverUsername" : "admin",
            "serverPassword" : "admin",
            "serverHost" : "localhost",
            "serverPort" : "4503",
            "deltaContent" : false,
            "pathConfigurations" :  [
                {
                    "path" : "/content/a/b",
                    "excludePaths" : ["d", "e/ab", "q/fg"],
                    "workflowConfigIds" : []
                }
            ]
        }
        """

    def expectedOutput = ["/content/a/b/d", "/content/a/b/e/ab", "/content/a/b/q/fg"]

    when:
    def actualOutput = GrabbitConfiguration.create(input)

    then:
    actualOutput instanceof GrabbitConfiguration
    actualOutput.pathConfigurations.first().excludePaths == expectedOutput
}
```

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -25,6 +25,125 @@ import spock.lang.Unroll
...
+
+    @Unroll
+    def "Should fail to return exclude paths"() {
```
jdigger: Not a very descriptive feature name. Maybe something more along the lines of `"throws exception for non-relative exclude paths"`

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -25,6 +25,125 @@ import spock.lang.Unroll
...
+           "pathConfigurations" :  [
+            {
+                "path" : "/content/a/b",
+                "excludePaths" : ["/a", "\\b", "./c"],
+                "workflowConfigIds" : []
+            }
+           ]
+        }
+        """ | [excludePaths: "relative paths provided cannot begin or end with /,./ or \\"]
+    }
```
jdigger: Better, that takes proper advantage of the `where:` capabilities (and shows that there's currently a bug):
```groovy
@Unroll
def "throws exception for non-relative exclude path: #excludePath"() {
    when:
    def input = """
    {
       "serverUsername" : "admin",
       "serverPassword" : "admin",
       "serverHost" : "localhost",
       "serverPort" : "4503",
       "deltaContent" : false,
       "pathConfigurations" :  [
        {
            "path" : "/content/a/b",
            "excludePaths" : ["${excludePath}"],
            "workflowConfigIds" : []
        }
       ]
    }
    """

    GrabbitConfiguration.create(input)

    then:
    GrabbitConfiguration.ConfigurationException exception = thrown()
    exception.errors == [excludePaths: "relative paths provided cannot begin or end with /,./ or \\"]

    where:
    excludePath << ["/a", "\\b", "./c"]
}
```

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -78,6 +79,23 @@ class GrabbitConfiguration {
         )
     }
 
+    private
+    static Collection<String> nonEmptyCollection(Collection<String> excludePaths, ConfigurationException.Builder errorBuilder) {
+        String prePattern = "^(/|\\./|\\|\\\\).*\$";
+        String postPattern = ".*(/|\\./|\\|\\\\)\$";
+        if (excludePaths == null || excludePaths.isEmpty())
+            return excludePaths
+        if (excludePaths.any({ String it -> (it == null) || (it.isEmpty()) })) {
```
jdigger: In Groovy, `it` is simply the name of the "unnamed" parameter. It's very weird to see it explicitly declared. (In fact, I was surprised that doing so wasn't a compile error.) Also, if the closure is the last (including "only") parameter to a function, it doesn't need to be inside method-call parens. Again, doing so is so unusual that it looks "really weird." :-)

This line can/should read simply
```groovy
if (excludePaths.any { it == null || it.isEmpty() }) {
```

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -123,10 +248,12 @@ class GrabbitConfigurationSpec extends Specification {
             "pathConfigurations" :  [
                 {
                     "path" : "/content/a/b",
+					"excludePaths" : [],
```
jdigger: What's with all the extra space? More hard tabs?

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
jdigger: It's not an actual "notation" -- it's being used like a comma in CSV.

grabbit/src/main/groovy/com/twcable/grabbit/server/services/impl/DefaultServerService.groovy
```diff
@@ -41,38 +41,41 @@ import javax.servlet.ServletOutputStream
...
-    void getContentForRootPath(@Nonnull String path, @Nullable String afterDateString, @Nonnull ServletOutputStream servletOutputStream) {
+    void getContentForRootPath(
+            @Nonnull String path,
+            @Nonnull String excludePaths,
+            @Nullable String afterDateString, @Nonnull ServletOutputStream servletOutputStream) {
 
-        if(path == null) throw new IllegalStateException("path == null")
-        if(servletOutputStream == null) throw new IllegalStateException("servletOutputStream == null")
+        if (path == null) throw new IllegalStateException("path == null")
+        if (excludePaths == null) throw new IllegalStateException("excludePaths == null")
```
jdigger: Why does "excludePaths" have to be required? In this context, if it's missing it could be treated as an empty String/collection.

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -78,6 +79,23 @@ class GrabbitConfiguration {
         )
     }
 
+    private
+    static Collection<String> nonEmptyCollection(Collection<String> excludePaths, ConfigurationException.Builder errorBuilder) {
+        String prePattern = "^(/|\\./|\\|\\\\).*\$";
+        String postPattern = ".*(/|\\./|\\|\\\\)\$";
+        if (excludePaths == null || excludePaths.isEmpty())
+            return excludePaths
```
jdigger: This seems really confusing, given the method name: you're returning a `null` or empty collection from a method called `nonEmptyCollection`.

jdigger: Where this is called, you're making sure this condition would never be triggered. Better would be to do the empty collection here.

So the calling code would be
```groovy
def excludePaths = nonEmptyCollection(config["excludePaths"], errorBuilder)
```
and then this condition would be
```groovy
if (excludePaths == null || excludePaths.isEmpty())
    return Collections.EMPTY_LIST
```

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -78,6 +79,23 @@ class GrabbitConfiguration {
         )
     }
 
+    private
+    static Collection<String> nonEmptyCollection(Collection<String> excludePaths, ConfigurationException.Builder errorBuilder) {
+        String prePattern = "^(/|\\./|\\|\\\\).*\$";
+        String postPattern = ".*(/|\\./|\\|\\\\)\$";
```
jdigger: Groovy allows "regex syntax" similar to perl/ruby/sed/vi/etc.
```groovy
String prePattern = /^(\/|\.\/|\|\\).*$/;
String postPattern = /^.*(\/|\.\/|\|\\)$/;
```

jdigger: BTW, if you didn't know, if you Alt-Enter inside the regex you can select "Check RegExp" and it will bring up a dialog to help test/build your regular expression. http://regexr.com/ and http://regexpal.com/ can be used for that too.

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -78,6 +79,23 @@ class GrabbitConfiguration {
...
+        if (excludePaths == null || excludePaths.isEmpty())
+            return excludePaths
+        if (excludePaths.any({ String it -> (it == null) || (it.isEmpty()) })) {
+            errorBuilder.add('excludePaths', 'contains null/empty')
+            return excludePaths
+        } else if (excludePaths.any { String it -> it.matches(prePattern) } || excludePaths.any { String it -> it.matches(postPattern) }) {
+            errorBuilder.add('excludePaths', 'relative paths provided cannot begin or end with /,./ or \\')
```
jdigger: Can't `./` just be a no-op like on the file system?

e.g.   `path: "/content/a", excludePaths: ["./b"]` -> `"/content/a/./b"` -> `"/content/a/b"`

jdigger: `"end with /,./ or \\"` <- Missing a space between the coma and the `./`

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -78,6 +79,23 @@ class GrabbitConfiguration {
...
+            return excludePaths
+        } else {
+            return excludePaths.collect({ String it -> it })
```
jdigger: What purpose does the `collect` provide here?

grabbit/src/functionalTest/groovy/com/twcable/grabbit/server/services/ServerServiceSpec.groovy
```diff
@@ -59,8 +59,7 @@ class ServerServiceSpec extends Specification {
 
         slingRepository = repository(fakeNodeBuilder).build()
 
-        syncServerService = new DefaultServerService(slingRepository: slingRepository,
-                configurableApplicationContext: configurableApplicationContext)
+        syncServerService = new DefaultServerService( slingRepository: slingRepository, configurableApplicationContext: configurableApplicationContext)
```
jdigger: Trivial: extra space   `DefaultServerService( slingRepository:`

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -78,6 +79,23 @@ class GrabbitConfiguration {
...
+    private
+    static Collection<String> nonEmptyCollection(Collection<String> excludePaths, ConfigurationException.Builder errorBuilder) {
+        String prePattern = /^(\/|\.\/|\\).*$/;
+        String postPattern = /^.*(\/|\.\/|\\)$/;
+        if (excludePaths == null || excludePaths.isEmpty())
+            return Collections.EMPTY_LIST
+        if (excludePaths.any({ it == null || it.isEmpty() })) {
+            errorBuilder.add('excludePaths', 'contains null/empty')
+            return excludePaths
+        } else if (excludePaths.any { it.matches(prePattern) } || excludePaths.any { it.matches(postPattern) } ) {
```
jdigger: Is there a logical difference between this and the simpler
```groovy
excludePaths.any { it.matches(prePattern) || it.matches(postPattern) }
```
?

jdigger:
The heart of the change is in `JcrNodesReader`, but there's no test that I can find: either explicitly (`JcrNodesReaderSpec`) or implicitly in an integration test.

Now that we know it can be configured correctly, I'd like to know it actually **_works_** correctly...  :broken_heart:

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
```diff
@@ -33,19 +32,39 @@ import javax.jcr.Node as JcrNode
 @SuppressWarnings("GrMethodMayBeStatic")
 class JcrNodesReader implements ItemReader<JcrNode> {
 
+    private String excludePaths
+    private String[] excludePathList
+    private static final String[] EMPTY_ARRAY = new String[0];
+
+    public void setExcludePaths(String excludePaths) {
+        this.excludePaths = excludePaths
```
jdigger: I don't see anywhere that this is referenced later. Why is it being set?

jdigger:
Avoid "raw" arrays when a collection type will work, like in `JcrNodesReader`.

* It's the wrong level of abstraction. There's nothing inherent about what the data means that says it's any more a sequence of memory addresses than that it's a List or Set or even just a Collection.
* Raw arrays don't play nicely with generics.
  * For type safety we're having to create `String[] EMPTY_ARRAY = new String[0]` and then `this.excludePathList = EMPTY_ARRAY`. If instead it were, for example, a `Collection<String>` then there would be no need to define a constant: you could let the Java compiler do the work and save memory: `this.excludePathList = Collections.EMPTY_LIST` (Then do a static import of `EMPTY_LIST` just to clean that up a little.)
* It's essentially a Java primitive, so it suffers from many of the same problems.
  * For example,  `excludePathList.any` is going to do "boxing" every time it's called to convert it from a `String[]` to an `Iterable<String>` so that it can iterate over it.
  * Because it's not an object, it doesn't have behavior, so you have to act "upon" it rather than "asking it to act". Groovy is having to do some magic to make it look like you can invoke a method (eg., `any(..)`) on the array. But in order to do so it has to box it first. And acting "upon" something rather than "messaging" it is violation number 1 of OO design.
    * Because you're acting upon it, it breaks any notion of encapsulation. While it's not a problem right here, say we needed to guarantee it's immutable. You'd have to track down every **_potential_** access of the array and make sure nothing could possibly change it, including making defensive copies. Or you could simply use Guava's ImmutableList and let polymorphism do its job. Or if you wanted it to be sorted in some specific way (e.g., alphabetically, from shortest to longest) because you know that the way it's going to be used would be more efficient if accessed that way: Use a sorted data-structure that pays the cost on write and reap the increase of speed on every read.
  * It won't be able to take advantage of advances in libraries that will come "for free" as the JDK advances, such as Java8 Streams.

Raw arrays are a reasonable thing in low-level programming, such as data serialization or for implementing higher-order data structures. But in general stay away from them. For this particular instance, none of this really matters a whole lot (the cost of the extra boxing is dwarfed by I/O), but on that same note there's no need to "prematurely optimize" to a raw array either.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
masroormohammed: This is similar to Workflow Configs I pass them from the xml , i.e., 

<bean id="serverJcrNodesReader" class="com.twcable.grabbit.server.batch.steps.jcrnodes.JcrNodesReader" scope="step">
    	<property name="excludePaths" value="#{jobParameters['excludePaths']}"/>
    </bean>

So that is the reason I have the setter

jdigger: I wasn't clear: I was referring to the field, not the method.

sagarsane:
Looks like the branch needs to be rebased with `upstream/master` properly.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
```diff
@@ -33,19 +33,38 @@ import javax.jcr.Node as JcrNode
...
+        while (true) {
+            if (nodeIterator.hasNext()) {
+                JcrNode node = nodeIterator.next()
+                if (!isPathInExcludedList(node.getPath())) {
+                    return node
+                }
+            } else {
+                return null
+            }
         }
```
jdigger: This complicated `while(true)` block can be replaced with
```groovy
return (JcrNode)nodeIterator.find { JcrNode node -> !isPathInExcludedList(node.path) }
```

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
```diff
@@ -33,19 +33,38 @@ import javax.jcr.Node as JcrNode
 @SuppressWarnings("GrMethodMayBeStatic")
 class JcrNodesReader implements ItemReader<JcrNode> {
 
+    private String excludePaths
```
jdigger: This field has been orphaned.

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -151,11 +169,20 @@ class GrabbitConfiguration {
...
         String path
+        Collection<String> excludePaths
         Collection<String> workflowConfigIds
 
+        void setPath(@Nonnull String path) {
+            this.path = path!=null && path.endsWith("/") ? path.substring(0, path.length() - 1): path
+        }
+
+        Collection<String> getAbsolutePaths(@Nonnull Collection<String> excludePaths) {
+            return excludePaths.collect { "${this.path}/${it}" }
```
jdigger: There's no invariant that `excludePaths` is non-null, so this can give an NPE...

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -151,11 +169,20 @@ class GrabbitConfiguration {
     @CompileStatic
     static class PathConfiguration {
         String path
+        Collection<String> excludePaths
         Collection<String> workflowConfigIds
 
+        void setPath(@Nonnull String path) {
+            this.path = path!=null && path.endsWith("/") ? path.substring(0, path.length() - 1): path
```
jdigger: The logic here allows `this.path` to be set to null. Is that okay?

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
jdigger: If `this.path` is null, then this will NPE

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
```diff
@@ -186,13 +187,14 @@ class ClientBatchJob {
...
-                "${USERNAME}"        : credentialsBuilder.username,
-                "${PASSWORD}"        : credentialsBuilder.password,
-                "${WORKFLOW_CONFIGS}": pathConfiguration.workflowConfigIds.join("|")
+                    "timestamp"           : System.currentTimeMillis() as String,
+                    "${PATH}"             : pathConfiguration.path,
+                    "${HOST}"             : serverBuilder.host,
+                    "${PORT}"             : serverBuilder.port,
+                    "${USERNAME}"         : credentialsBuilder.username,
+                    "${PASSWORD}"         : credentialsBuilder.password,
+                    "${EXCLUDE_PATHS}"    : pathConfiguration.excludePaths.join("<sp>"),
```
sagarsane: This may have been already discussed, but it seems odd that the join is using `<sp>` and not like `|` as done below.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesReader.groovy
```diff
@@ -33,19 +33,37 @@ import javax.jcr.Node as JcrNode
...
+    private Collection<String> excludePathList
+
+    public void setExcludePaths(String excludePaths) {
+        if (excludePaths != null && !excludePaths.isEmpty()) {
+            this.excludePathList = excludePaths.split("<sp>") as Collection<String>
```
jdigger: Per [the JCR spec for path naming](http://www.day.com/specs/jcr/2.0/3_Repository_Model.html#3.2.4%20Naming%20Restrictions), "&lt;sp>" is perfectly legal (if unlikely) to be part of a path. You might want to use colon instead, since it's disallowed.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
jdigger: The problem with `|` is that it's valid to be in a path (whereas it's not valid to be in a workflow config id, so it can be used as a delimiter for those). Of course, so can `<sp>`, but that gets to [my other comment..](https://github.com/TWCable/grabbit/pull/29#discussion_r31847833)

masroormohammed:
@jdigger ready for another round of review

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/ServerBatchJob.groovy
```diff
@@ -89,16 +90,19 @@ class ServerBatchJob {
...
 
-
-        ContentAfterDateBuilder andPath(String path) {
-            if (path == null) throw new IllegalArgumentException("path == null")
+        ContentAfterDateBuilder andPath(String path, String excludePaths) {
+            if(path == null) throw new IllegalArgumentException("path == null")
             this.path = path
+            if(excludePaths != null && !excludePaths.isEmpty()) {
+                this.excludePaths = excludePaths
+            }
```
jdigger:
```groovy
this.excludePaths = (excludePaths == null || excludePaths.isEmpty()) ? "" : excludePaths
```

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
```diff
@@ -186,13 +187,14 @@ class ClientBatchJob {
...
-                "${PASSWORD}"        : credentialsBuilder.password,
-                "${WORKFLOW_CONFIGS}": pathConfiguration.workflowConfigIds.join("|")
+                    "timestamp"           : System.currentTimeMillis() as String,
+                    "${PATH}"             : pathConfiguration.path,
+                    "${HOST}"             : serverBuilder.host,
+                    "${PORT}"             : serverBuilder.port,
+                    "${USERNAME}"         : credentialsBuilder.username,
+                    "${PASSWORD}"         : credentialsBuilder.password,
+                    "${EXCLUDE_PATHS}"    : pathConfiguration.excludePaths.join("*"),
+                    "${WORKFLOW_CONFIGS}" : pathConfiguration.workflowConfigIds.join("|")
```
jdigger: Minor, but why is the formatting changed for all these?

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -151,11 +170,23 @@ class GrabbitConfiguration {
     @CompileStatic
     static class PathConfiguration {
         String path
+        Collection<String> excludePaths
         Collection<String> workflowConfigIds
 
+        void setPath(@Nullable String path) {
+            this.path = path!=null && path.endsWith("/") ? path.substring(0, path.length() - 1): path
```
jdigger: Really should use parentheses here for clarity.

jdigger: More Groovy niceness:
```groovy
path.substring(0, path.length() - 1)
```
->
```groovy
path[0..-2]
```
The negative number says "Start from the right and start going left, inclusive." Or think about it like `string.length() - n` inclusively. Since the end number of the range is inclusive, `-1` is the last character of the string.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/ServerBatchJob.groovy
```diff
@@ -142,13 +146,14 @@ class ServerBatchJob {
...
-                    .toJobParameters(),
-                (JobLauncher)configurationBuilder.configAppContext.getBean("serverJobLauncher", JobLauncher)
+                    (Job) configurationBuilder.configAppContext.getBean("serverJob", Job),
+                    new JobParametersBuilder()
+                            .addLong("timestamp", System.currentTimeMillis())
+                            .addString(PATH, afterDateBuilder.pathBuilder.path)
+                            .addString(EXCLUDE_PATHS, afterDateBuilder.pathBuilder.excludePaths)
+                            .addString(CONTENT_AFTER_DATE, afterDateBuilder.contentAfterDate)
+                            .toJobParameters(),
+                    (JobLauncher) configurationBuilder.configAppContext.getBean("serverJobLauncher" ,JobLauncher)
```
jdigger: Minor, but why is the formatting changed for all these?

grabbit/src/test/groovy/com/twcable/grabbit/server/services/JcrContentExclusionIteratorSpec.groovy
```diff
@@ -0,0 +1,81 @@
...
+        JcrNode rootNode = fakeNodeBuilder.build()
+
+        when:
+        final Iterator<JcrNode> nodeIterator = new JcrContentExclusionIterator(new JcrContentRecursiveIterator(rootNode), "page/jcr:content")
+        final JcrNode root = nodeIterator.next()
+        final JcrNode firstChild = nodeIterator.next()
+        then:
+        root.name == "page"
+        firstChild.name != "jcr:content"
+
```
jdigger:
```groovy
when:
final nodeIterator = new JcrContentExclusionIterator(new JcrContentRecursiveIterator(rootNode), "page/jcr:content")
final root = nodeIterator.next()
def childNodes = nodeIterator.collect()

then:
root.name == "page"
childNodes.size() == 2
childNodes.find { JcrNode node -> node.name == "jcr:content" } == null
```
The advantage to this approach is that while we are relying on the iterator being either depth-first or breadth-first, there's no other assumption about the ordering. (Such as alphabetical by name, insertion order, grouped by type, creation time, etc.)

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -24,6 +24,143 @@ import spock.lang.Unroll
...
+                        "path" : "/content/dam/d/images",
+                        "excludePaths" :["f",null],
+                        "workflowConfigIds" :
+                        [
+                            "something"
+                        ]
+                    }
+                ]
+            }
+            """ | [excludePaths: "contains null/empty"]
```
jdigger: When there's only one set of inputs, `where:` is adding complexity with no benefit.

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -24,6 +24,143 @@ import spock.lang.Unroll
...
+
+        then:
+        GrabbitConfiguration.ConfigurationException exception = thrown()
+        exception.errors == [excludePaths: "relative paths provided cannot begin or end with /, ./ or \\"]
+
+        where:
+        excludePath << ["/a", "\\\\b" , "./c"]
+    }
+
+    def "generates correct exclusions regardless of trailing slash for path: #excludePath"() {
```
jdigger: Need an `@Unroll` to unroll `excludePath` in the description.

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -24,6 +24,143 @@ import spock.lang.Unroll
...
+        then:
+        actualOutput instanceof GrabbitConfiguration
+        actualOutput.pathConfigurations.first().excludePaths == expectedOutput
+
+        where:
+        path << ["/content/a/b", "/content/a/b/"]
+    }
+
+    @Unroll
+    def "Should fail to process exclude paths"() {
```
jdigger: No need for `@Unroll` when there's no parameters in the feature description.

grabbit/src/main/groovy/com/twcable/grabbit/servlets/GrabbitServlet.groovy
```diff
@@ -56,15 +56,14 @@ class GrabbitServlet extends SlingAllMethodsServlet {
 
         if (path) { //Server Get Request
             final decodedPath = URLDecoder.decode(path, "utf-8")
+            final decodedExcludePaths = URLDecoder.decode(request.getParameter("excludePaths") ?: "", "utf-8")
```
jdigger: Instead of worrying about using special delimiters for sending between client and server, just use what's built into the servlet spec: [`getParameterValues(..)`](https://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getParameterValues(java.lang.String))

That means that you can pass around `Collection<String>` through the layers of the API instead of "a "`String`" but it's actually a collection of strings that hasn't been separated yet".

We'll still need to use a delimiter for the Batch Job parameters, but if it's later discovered that the chosen delimiter isn't appropriate, then it's an internal detail/fix rather than a part of the communication protocol between client and server. (Remember, ideally client and server should be able to evolve independently so that **_every_** upper environment doesn't have to change at exactly the same rate as **_every_** lower environment.)

grabbit/src/main/groovy/com/twcable/grabbit/server/services/JcrContentExclusionIterator.groovy
```diff
@@ -0,0 +1,62 @@
...
+    private Collection<String> excludePathList
+
+    public JcrContentExclusionIterator(Iterator<JcrNode> nodeIterator, String excludePaths) {
+        this.nodeIterator = nodeIterator
+        if (excludePaths != null && !excludePaths.isEmpty()) {
+            this.excludePathList = excludePaths.split("\\*") as Collection<String>
+        } else {
+            this.excludePathList = Collections.EMPTY_LIST
+        }
+    }
```
jdigger: If you decode the exclude paths at the servlet side, then this becomes simply
```groovy
public JcrContentExclusionIterator(Iterator<JcrNode> nodeIterator, Collection<String> excludePaths) {
    this.nodeIterator = nodeIterator
    this.excludePathList = (excludePaths == null) ? (Collection<String>)Collections.EMPTY_LIST : excludePaths
}
```

grabbit/src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -90,6 +232,7 @@ class GrabbitConfigurationSpec extends Specification {
                     },
                     {
                         "path" : "/content/dam/d/images",
+                        "excludePaths" : [],
```
jdigger: Why are all these
```groovy
"excludePaths" : [],
```
lines here? Doesn't the JSON work if it's omitted? It seems like it's just added a lot of cruft/complexity to the spec with no benefit.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/ServerBatchJob.groovy
jdigger: Of course this line becomes just a little different with `Collection<String>`...

README.md
```diff
@@ -72,10 +72,16 @@ A `json` configuration file of following format is used to configure Grabbit.
...
+            "path" : "/content/someContent",
+            "excludePaths" : "someContent/someExcludeContent"
         },
         {
             "path" : "/content/dam/someDamContent",
+            "excludePaths":
+                [
+                    "someContent/someExcludeContent",
+                    "someContent/someOtherExcludeContent"
+                ],
```
jdigger: Where is the description of what "excludePaths" means and what rules there are around it?

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -81,6 +83,23 @@ class GrabbitConfiguration {
...
+        if (excludePaths.any({ it == null || it.isEmpty() })) {
+            errorBuilder.add('excludePaths', 'contains null/empty')
+            return excludePaths
+        } else if (excludePaths.any { it.matches(prePattern) || it.matches(postPattern) } ) {
+            errorBuilder.add('excludePaths', 'relative paths provided cannot begin or end with /, ./ or \\')
+            return excludePaths
+        } else {
+            return excludePaths
+        }
+    }
```
jdigger: Minor, but slightly more efficient:
```groovy
private static final Pattern prePattern = Pattern.compile(/^(\/|\.\/|\\).*$/)
private static final Pattern postPattern = Pattern.compile(/^.*(\/|\.\/|\\)$/)

private static Collection<String> nonEmptyCollection(Collection<String> excludePaths,
                                                     ConfigurationException.Builder errorBuilder) {
    if (excludePaths == null || excludePaths.isEmpty())
        return Collections.EMPTY_LIST

    if (excludePaths.any({ it == null || it.isEmpty() }))
        errorBuilder.add('excludePaths', 'contains null/empty')
    else if (excludePaths.any { it.matches(prePattern) || it.matches(postPattern) })
        errorBuilder.add('excludePaths', 'relative paths provided cannot begin or end with /, ./ or \\')

    return excludePaths
}
```

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJobExecutionListener.groovy
```diff
@@ -113,6 +114,8 @@ class ClientBatchJobExecutionListener implements JobExecutionListener {
      */
     private HttpResponse doRequest(JobParameters jobParameters) {
         final String path = jobParameters.getString(ClientBatchJob.PATH)
+        final String excludePathParam = jobParameters.getString(ClientBatchJob.EXCLUDE_PATHS)
+        final Collection<String> excludePaths =  (excludePathParam!=null) ? excludePathParam.split("\\*") as Collection<String> : (Collection<String>)Collections.EMPTY_LIST
```
jdigger: A little cleaner:
```groovy
final excludePaths = (excludePathParam != null ? excludePathParam.split(/\*/) : Collections.EMPTY_LIST) as Collection<String>
```

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJobExecutionListener.groovy
```diff
@@ -121,10 +124,16 @@ class ClientBatchJobExecutionListener implements JobExecutionListener {
 
         final String encodedContentAfterDate = URLEncoder.encode(contentAfterDate, 'utf-8')
         final String encodedPath = URLEncoder.encode(path, 'utf-8')
-        final grabPath = "http://${host}:${port}/grabbit/job?path=${encodedPath}&after=${encodedContentAfterDate}"
+
+        URIBuilder uriBuilder = new URIBuilder("http://${host}:${port}/grabbit/job")
```
jdigger: Clearer, and more efficient:
```groovy
URIBuilder uriBuilder = new URIBuilder(scheme: "http", host: host, port: port as Integer, path: "/grabbit/job")
```

grabbit/src/main/groovy/com/twcable/grabbit/server/services/impl/DefaultServerService.groovy
```diff
@@ -71,11 +75,14 @@ class DefaultServerService implements ServerService {
...
+            //Iterator wrapper for excludePaths exclusions
+            nodeIterator = new JcrContentExclusionIterator(nodeIterator, excludePaths)
+
             ServerBatchJob batchJob = new ServerBatchJob.ConfigurationBuilder(configurableApplicationContext)
-                .andConfiguration(new NamespaceHelper(session).namespaces.iterator(), nodeIterator, servletOutputStream)
-                .andPath(path)
-                .andContentAfterDate(afterDateString)
-                .build()
+                    .andConfiguration(new NamespaceHelper(session).namespaces.iterator(), nodeIterator, servletOutputStream)
+                    .andPath(path, excludePaths.join("*"))
```
jdigger: The `excludePaths` should be passed in as a `Collection<String>` and let the Batch Job Builder deal with "serializing" it to a `String` for better cohesion.

grabbit/src/main/groovy/com/twcable/grabbit/server/services/impl/DefaultServerService.groovy
```diff
@@ -71,11 +75,14 @@ class DefaultServerService implements ServerService {
...
 
+            //Iterator wrapper for excludePaths exclusions
+            nodeIterator = new JcrContentExclusionIterator(nodeIterator, excludePaths)
+
             ServerBatchJob batchJob = new ServerBatchJob.ConfigurationBuilder(configurableApplicationContext)
-                .andConfiguration(new NamespaceHelper(session).namespaces.iterator(), nodeIterator, servletOutputStream)
-                .andPath(path)
-                .andContentAfterDate(afterDateString)
-                .build()
+                    .andConfiguration(new NamespaceHelper(session).namespaces.iterator(), nodeIterator, servletOutputStream)
```
jdigger: Why are these lines indented more?

grabbit/src/main/groovy/com/twcable/grabbit/servlets/GrabbitServlet.groovy
```diff
@@ -53,18 +53,21 @@ class GrabbitServlet extends SlingAllMethodsServlet {
...
     void doGet(SlingHttpServletRequest request, SlingHttpServletResponse response) {
         final path = request.getParameter("path")
+        final excludePaths = request.getParameterValues("excludePath")
 
         if (path) { //Server Get Request
             final decodedPath = URLDecoder.decode(path, "utf-8")
+            for(int i=0; i<excludePaths.length; i++) {
+                excludePaths[i] = URLDecoder.decode(excludePaths[i], 'UTF-8')
+            }
+            final decodedExcludePaths = excludePaths as Collection<String>
```
jdigger:
```groovy
final decodedExcludePaths = excludePaths.collect { String ep -> URLDecoder.decode(ep, 'UTF-8') }
```

jdigger:
Just a few minor items left. CR +1 :+1:

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJobExecutionListener.groovy
sagarsane: :+1: yeah much cleaner.

masroormohammed:
@jdigger modified the Readme File and the minor items. Was able to test it properly after the version change to 4.3.6 for http.utils,  If the Readme changes look fine to you I'll go ahead and Squash the commits.

README.md
```diff
@@ -62,6 +62,37 @@ _For both Windows and Mac : To verify that installation was successful, `protoc
...
 
+#### Exclude Paths: ####
+
+This allows the user to blacklist a certain set of paths for a given path. As a result those paths would be eliminated while grabbing content, there are certain rules that need to be followed,
+
+1. Exclude Paths only accepts relative paths and not absolute paths.
+
+    For example,
+
+    Let's say given path : `"/content/someContent"`  and you would like to remove `"/content/someContent/someOtherContent/pdfs"`
```
sagarsane: `... would like to exclude` ..?

masroormohammed: yup caught that, was waiting for any more changes for the readme file before I push the final commit.

README.md
```diff
@@ -62,6 +62,37 @@ _For both Windows and Mac : To verify that installation was successful, `protoc
...
+        ]
+    }
+    ```
+    You could also provide multiple exclusions for a single path(similar to workflow config ids)
+
+2. Following are INVALID characters to be provided as part of Exclusion Paths,
+
+    relative paths provided cannot begin or end with `/, ./ or \`
+
+3. Exclude Paths is not a mandatory field so it can be omitted from path configuration as well.
```
sagarsane: May be we can highlight this point some how (minor .. just a suggestion)

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJobExecutionListener.groovy
```diff
@@ -121,10 +124,16 @@ class ClientBatchJobExecutionListener implements JobExecutionListener {
...
         final String encodedContentAfterDate = URLEncoder.encode(contentAfterDate, 'utf-8')
         final String encodedPath = URLEncoder.encode(path, 'utf-8')
-        final grabPath = "http://${host}:${port}/grabbit/job?path=${encodedPath}&after=${encodedContentAfterDate}"
+
+        URIBuilder uriBuilder = new URIBuilder(scheme: "http", host: host, port: port as Integer, path: "/grabbit/job")
+        uriBuilder.addParameter("path", encodedPath)
+        uriBuilder.addParameter("after", encodedContentAfterDate)
+        for(String excludePath : excludePaths) {
+            uriBuilder.addParameter("excludePath", URLEncoder.encode(excludePath, 'UTF-8'))
+        }
```
sagarsane: @jdigger @masroormohammed this for loop will basically add ... `?excludedPath=a&excludedPath=b&excludedPath=c` and so on. right? At first this seemed odd, but then I saw `GrabbitServlet` has `request.getParameterValues()` which then makes sense.

Just wanted to confirm ...

masroormohammed: Yup that's what we decided to go with, the underlying structure takes away the issue of having the unique delimiter being the bottleneck of the communication between the client and the server... we would still need the delimiter for the job parameter though and changes to that would be minimal incase we need to change the delimiter(doesn't affect the communication between client and server)

sagarsane: Makes sense :)

README.md
```diff
@@ -62,6 +62,37 @@ _For both Windows and Mac : To verify that installation was successful, `protoc
...
+    }
+    ```
+    You could also provide multiple exclusions for a single path(similar to workflow config ids)
+
+2. Following are INVALID characters to be provided as part of Exclusion Paths,
+
+    relative paths provided cannot begin or end with `/, ./ or \`
+
+3. `Exclude Paths is not a mandatory field` so it can be omitted from path configuration as well.
+
```
jdigger: Better:

### Config Format

A `json` object is used by the PUT to the client to tell it what to request from the server. An example configuration:

```json
{
    "serverUsername" : "<username>",
    "serverPassword" : "<password>",
    "serverHost" : "some.other.server",
    "serverPort" : "4502",
    "pathConfigurations" :  [
        {
            "path" : "/content/someContent",
        },
        {
            "path" : "/content/someContent",
            "excludePaths" :
            [
                "someOtherContent/someExcludeContent"
            ]
        },
        {
            "path" : "/content/dam/someDamContent",
            "excludePaths":
                [
                    "someContent/someExcludeContent",
                    "someContent/someOtherExcludeContent"
                ],
            "workflowConfigIds" :
                [
                    "/etc/workflow/launcher/config/update_asset_mod",
                    "/etc/workflow/launcher/config/update_asset_create",
                    "/etc/workflow/launcher/config/dam_xmp_nested_writeback",
                    "/etc/workflow/launcher/config/dam_xmp_writeback"
                ]
        }
    ]
}
```


#### Required fields

* **serverHost**: The server that the client should get its content from.
* **serverPort**: The port to connect to on the server that the client should use.
* **serverUsername**: The username the client should use to authenticate against the server.
* **serverPassword**: The password the client should use to authenticate against the server.
* **pathConfigurations**: The list of paths and their options to pull from the server.
  * **path**: The path to recursively grab content from.


#### Optional fields

Under "path configurations"

* **excludePaths**: This allows excluding specific subpaths from what will be retrieved from the parent path. See more detail below.
* **workflowConfigIds**: Before the client retrieves content for the path from the server, it will make sure that the specified workflows are disabled. They will be re-enabled when all content specifying that workflow has finished copying. (Grabbit handles the situation of multiple paths specifying "overlapping" workflows.) This is particularly useful for areas like the DAM where a number of relatively expensive workflows will just "redo" what is already being copied.


#### Exclude Paths

Exclude Paths allow the user to exclude a certain set of subpaths for a given path while grabbing content. They can only be **relative** to the "path".

For example, let's say you have
```json
{ "path" : "/content/someContent" }
```
and you would like to exclude `/content/someContent/someOtherContent/pdfs`

Valid:
```json
{
    "path" : "/content/someContent",
    "excludePaths" :
    [
        "someOtherContent/pdfs"
    ]
}
```
Invalid:
```json
{
    "path" : "/content/someContent",
    "excludePaths" :
    [
        "/content/someContent/someOtherContent/pdfs",
        "/someOtherContent/pdfs"
        "./someOtherContent/pdfs"
    ]
}
```

masroormohammed:
@jdigger Changes to readme are done and commits are squashed.

PR #32, opened 2015-06-08
Add mixins before setting properties

sagarsane:
This should prevent "no matching property found" (`ConstraintViolationException`) exception. It seems like `node.addMixin()` should be reflected as soon as its executed (hence, I didn't call `session.save()`)

@jazeren1 @jdigger


grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
```diff
@@ -118,16 +118,15 @@ class JcrNodesWriter implements ItemWriter<NodeProtos.Node>, ItemWriteListener {
...
-                if (JCR_PRIMARYTYPE != protoProperty.name && (protoProperty.hasValue() || protoProperty.hasValues())) {
-                    if (protoProperty.name == JCR_MIXINTYPES) {
-                        addMixins(protoProperty, currentNode)
-                    }
-                    else {
-                        final String propertyPrefix = protoProperty.name.split(":")[0]
-                        log.debug "Current node ${protoProperty.name} prefix : ${propertyPrefix}"
-                        addProperty(protoProperty, currentNode)
-                    }
+                if( !(protoProperty.name in [JCR_PRIMARYTYPE, JCR_MIXINTYPES]) && (protoProperty.hasValue() || protoProperty.hasValues())) {
```
sagarsane: This HAS gotten quite complex though ....

jdigger:
Is there a test that validates this?

b-Johnson:
Still seeing an error after building this PR.

08.06.2015 11:57:06.609 *DEBUG* [127.0.0.1 [1433789826599] GET /grabbit/job/all.json HTTP/1.1] com.twcable.grabbit.spring.batch.repository.JcrStepExecutionDao Mapped StepExecution : StepExecution: id=4285251050827383226, version=15, name=clientJcrNodes, status=FAILED, exitStatus=FAILED, readCount=700, filterCount=0, writeCount=600 readSkipCount=0, writeSkipCount=0, processSkipCount=0, commitCount=6, rollbackCount=1, exitDescription=javax.jcr.nodetype.ConstraintViolationException: No matching property definition: dam:folderThumbnailPaths = [/content/dam/sfu/myDepartment/Screen shot 2010-09-22 at 12.58.33 PM.png, /content/dam/sfu/waterfnt/New Westminster Harbour 1936_v2.jpg, /content/dam/sfu/waterfnt/nwest_waterfront1.png]
	at org.apache.jackrabbit.oak.jcr.delegate.NodeDelegate.setProperty(NodeDelegate.java:522)
	at org.apache.jackrabbit.oak.jcr.session.NodeImpl$36.perform(NodeImpl.java:1343)
	at org.apache.jackrabbit.oak.jcr.session.NodeImpl$36.perform(NodeImpl.java:1332)
	at org.apache.jackrabbit.oak.jcr.delegate.SessionDelegate.perform(SessionDelegate.java:293)
	at org.apache.jackrabbit.oak.jcr.session.ItemImpl.perform(ItemImpl.java:113.....

sagarsane:
Opened an issue on Jackalope to implement more API methods (including that for node mixins) https://github.com/TWCable/jackalope/issues/6 as that's preventing testing this code change.

sagarsane:
@b-Johnson , the exception above seems to be a "DEBUG" statement of the Grabbit Status API. Are you still seeing the same exception on the Grabbit Client when actually copying the content at `/content/dam/sfy/myDepartMent` ?

b-Johnson:
@sagarsane 
Yes, Here is the error from the error.log
08.06.2015 11:56:53.817 *ERROR* [clientJobLauncherTaskExecutor-2] org.springframework.batch.core.step.AbstractStep Encountered an error executing step clientJcrNodes in job clientJob
javax.jcr.nodetype.ConstraintViolationException: No matching property definition: dam:folderThumbnailPaths = [/content/dam/sfu/myDepartment/Screen shot 2010-09-22 at 12.58.33 PM.png, /content/dam/sfu/waterfnt/New Westminster Harbour 1936_v2.jpg, /content/dam/sfu/waterfnt/nwest_waterfront1.png]
	at org.apache.jackrabbit.oak.jcr.delegate.NodeDelegate.setProperty(NodeDelegate.java:522)
	at org.apache.jackrabbit.oak.jcr.session.NodeImpl$36.perform(NodeImpl.java:1343)
	at org.apache.jackrabbit.oak.jcr.session.NodeImpl$36.perform(NodeImpl.java:1332)
	at org.apache.jackrabbit.oak.jcr.delegate.SessionDelegate.perform(SessionDelegate.java:293)....

jdigger:
"It seems like node.addMixin() should be reflected as soon as its executed"  According to the spec, that's undefined -- it can happen right away or at save. However it should be reflected within the current session just fine, so other operations within that session should see it.

sagarsane:
Right. Yes that makes sense and as the operation is on the `current node` in the session, unless we have 2 config paths with overlapping trees, this should not be a problem I think.

jbornemann:
Started a bit of clean-up in response to one of Sagar's comments.  Still not complete, but wanted to get some eyes on.  Need to write another unit test.

jbornemann:
Should help the nt:hierarchy enhancement roll a bit more smoothly too

jbornemann:
@jdigger @sagarsane ready for review

grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
```diff
@@ -0,0 +1,56 @@
...
+@CompileStatic
+@Slf4j
+class JCRNodeDecorator {
+
+    @Delegate
+    JCRNode innerNode
+
+
+    JCRNodeDecorator(@Nonnull JCRNode node) {
+        if(!node) throw new IllegalArgumentException("node must not be null!")
```
sikarwarvimal: You are throwing the exception here but where it has been handled? In case of exception what will happen to the flow?

grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
```diff
@@ -0,0 +1,56 @@
...
+        this.innerNode = node
+    }
+
+
+    void setLastModified() {
+        final lastModified = new DateValue(Calendar.instance)
+        try {
+            //Need to check if jcr:lastModified can be added to the current node via its NodeType definition
+            //as it cannot be added to all the nodes
+            if (innerNode.primaryNodeType.canSetProperty(JCR_LASTMODIFIED, lastModified)) {
```
sikarwarvimal: If `innerNode.primaryNodeType` is null?

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
```diff
@@ -0,0 +1,90 @@
...
+    @Delegate
+    ProtoNode innerProtoNode
+
+    Collection<ProtoPropertyDecorator> protoProperties
+
+
+    ProtoNodeDecorator(@Nonnull ProtoNode node) {
+        if(!node) throw new IllegalArgumentException("node must not be null!")
+        this.innerProtoNode = node
+        this.protoProperties = node.properties.propertyList.collect { new ProtoPropertyDecorator(it) }
```
sikarwarvimal: need null check?

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+
+    /**
+     * If a property can be added as a mixin, adds it to the given node
+     * @param property
+     * @param node
+     */
+    private static void addMixins(ProtoPropertyDecorator property, JCRNode node) {
+        property.values.valueList.each { ProtoValue value ->
```
sikarwarvimal: need null check `property.values.valueList`?

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
```diff
@@ -74,9 +69,9 @@ class JcrNodesWriter implements ItemWriter<NodeProtos.Node>, ItemWriteListener {
 
 
     @Override
-    void write(List<? extends NodeProtos.Node> nodeProtos) throws Exception {
+    void write(List<? extends ProtoNode> nodeProtos) throws Exception {
         Session session = theSession()
-        for (NodeProtos.Node nodeProto : nodeProtos) {
+        for (ProtoNode nodeProto : nodeProtos) {
             writeToJcr(nodeProto, session)
         }
```
sikarwarvimal: A bit groovified `nodeProtos.each{writeToJcr(it, session)}`

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
```diff
@@ -0,0 +1,92 @@
...
+        return innerProtoProperty.values.valueList.collect { ProtoValue protoValue -> getJCRValueFromProtoValue(protoValue) } as Value[]
+    }
+
+
+    private Value getJCRValueFromProtoValue(ProtoValue value) throws ValueFormatException {
+
+        final valueFactory = ValueFactoryImpl.getInstance()
+
+        if (innerProtoProperty.type == PropertyType.BINARY) {
+            final binary = valueFactory.createBinary(new ByteArrayInputStream(value.bytesValue.toByteArray()))
```
sikarwarvimal: if null `value.bytesValue`?

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
```diff
@@ -0,0 +1,92 @@
...
+    private Value[] getPropertyValues() throws ValueFormatException {
+        return innerProtoProperty.values.valueList.collect { ProtoValue protoValue -> getJCRValueFromProtoValue(protoValue) } as Value[]
+    }
+
+
+    private Value getJCRValueFromProtoValue(ProtoValue value) throws ValueFormatException {
+
+        final valueFactory = ValueFactoryImpl.getInstance()
+
+        if (innerProtoProperty.type == PropertyType.BINARY) {
```
sikarwarvimal: Will prefer `switch` instead of `if` mostly for increased readability and in some cases slightly better performance.
http://www.blackwasp.co.uk/SpeedTestIfElseSwitch.aspx

sikarwarvimal: Thought the evaluation id on .Net platform but Java also provide a bit better performance.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+
+    private Value getPropertyValue() throws ValueFormatException {
+        getJCRValueFromProtoValue(innerProtoProperty.getValue())
+    }
+
+
+    private Value[] getPropertyValues() throws ValueFormatException {
+        return innerProtoProperty.values.valueList.collect { ProtoValue protoValue -> getJCRValueFromProtoValue(protoValue) } as Value[]
```
sikarwarvimal: null check `innerProtoProperty.values.valueList`?

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
```diff
@@ -0,0 +1,90 @@
...
+            if (node.canAddMixin(value.stringValue)) {
+                node.addMixin(value.stringValue)
+                log.info "Added mixin ${value.stringValue} for : ${node.name}."
```
sikarwarvimal: debug label log?

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
```diff
@@ -0,0 +1,90 @@
...
+
+    ProtoNodeDecorator(@Nonnull ProtoNode node) {
+        if(!node) throw new IllegalArgumentException("node must not be null!")
+        this.innerProtoNode = node
+        this.protoProperties = node.properties.propertyList.collect { new ProtoPropertyDecorator(it) }
+    }
+
+
+    String getPrimaryType() {
+        protoProperties.find { it.isPrimaryType() }.value.stringValue
```
sikarwarvimal: null check `.value.stringValue`?

sikarwarvimal:
Any suggested steps for runtime tests?

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
```diff
@@ -0,0 +1,92 @@
...
+            if (ex.message.contains("Multivalued property can not be set to a single value")) {
+                //If this is the exception, that means that a property with the name already exists
+                final Property currentProperty = node.getProperty(innerProtoProperty.name)
+                if (currentProperty.multiple) {
+                    final Value[] values = [getPropertyValue()]
+                    node.setProperty(innerProtoProperty.name, values, innerProtoProperty.type)
+                }
+            }
+            else if (ex.message.contains("Single-valued property can not be set to an array of values")) {
+                node.setProperty(innerProtoProperty.name, getPropertyValues().first(), innerProtoProperty.type)
```
sikarwarvimal: I understood the above logic but wondering if there is a better way to handle single/multiple values rather than relying the exception message?

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy
```diff
@@ -0,0 +1,210 @@
...
+                                                    .setStringValue("somevalue")
+                                                    .build()
+                                            )
+                                            .build()
+
+        final propertyDecorator = new ProtoPropertyDecorator(someProperty)
+        propertyDecorator.writeToNode(mockNode)
+
+        then:
+        1 * mockNode.setProperty("someproperty", new StringValue("somevalue"), PropertyType.STRING)
```
sikarwarvimal: Just to avoid hardcoded values at multiple places:
```
when:
        PropertyProto someProperty = PropertyProto.newBuilder()
                                            .setName(property)
                                            .setType(PropertyType.STRING)
                                            .setValue(
                                                ProtoValue.newBuilder()
                                                    .setStringValue(value)
                                                    .build()
                                            )
                                            .build()

        final propertyDecorator = new ProtoPropertyDecorator(someProperty)
        propertyDecorator.writeToNode(mockNode)

        then:
        1 * mockNode.setProperty(property, new StringValue(value), PropertyType.STRING)

        where:
        property        | value
        "someproperty"  | "somevalue"
```

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy
```diff
@@ -0,0 +1,210 @@
...
+                                                      )
+                                                      .build()
+                                              )
+                                              .build()
+
+        final propertyDecorator = new ProtoPropertyDecorator(someMultiProperty)
+        propertyDecorator.writeToNode(mockNode)
+
+        then:
+        1 * mockNode.setProperty("somemultivalueproperty", [new StringValue("value1"), new StringValue("value2")] as Value[], PropertyType.STRING)
```
sikarwarvimal: Same could be here

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy
```diff
@@ -0,0 +1,210 @@
...
+                isMultiple() >> true
+            }
+        }
+
+        final propertyDecorator = new ProtoPropertyDecorator(someProperty)
+        propertyDecorator.writeToNode(mockNode)
+
+        then:
+        notThrown(ValueFormatException)
+        1 * mockNode.setProperty("somename", [new StringValue("somevalue")] as Value[], PropertyType.STRING)
```
sikarwarvimal: Here too

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+
+    def "A multi-value property that is attempted to be written to a single-value property can recover, and be written"() {
```
sikarwarvimal: Getting below exception for this:

```
Too few invocations for:

1 * mockNode.setProperty("somemultivalueproperty", new StringValue("value1"), PropertyType.STRING)   (0 invocations)

Unmatched invocations (ordered by similarity):

1 * mockNode.getProperty('somemultivalueproperty')


	at org.spockframework.mock.runtime.InteractionScope.verifyInteractions(InteractionScope.java:78)
	at org.spockframework.mock.runtime.MockController.leaveScope(MockController.java:76)
	at com.twcable.grabbit.jcr.ProtoPropertyDecoratorSpec.A multi-value property that is attempted to be written to a single-value property can recover, and be written(ProtoPropertyDecoratorSpec.groovy:160)
```

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy
```diff
@@ -0,0 +1,210 @@
...
+
+        final propertyDecorator = new ProtoPropertyDecorator(someProperty)
+        propertyDecorator.writeToNode(mockNode)
+
+        then:
+        1 * mockNode.setProperty("someproperty", new StringValue("somevalue"), PropertyType.STRING)
+    }
+
+
+    def "A multi-value property can be written correctly to a node that expects a multi-value property"() {
```
sikarwarvimal: Getting below exception for this:
```
java.lang.NullPointerException
	at com.twcable.grabbit.jcr.ProtoPropertyDecorator.writeToNode(ProtoPropertyDecorator.groovy:38)
	at com.twcable.grabbit.jcr.ProtoPropertyDecoratorSpec.A multi-value property can be written correctly to a node that expects a multi-value property(ProtoPropertyDecoratorSpec.groovy:99)
```

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoNodeDecoratorSpec.groovy
```diff
@@ -0,0 +1,127 @@
...
+import static org.apache.jackrabbit.JcrConstants.JCR_MIXINTYPES
+import static org.apache.jackrabbit.JcrConstants.JCR_PRIMARYTYPE
+
+@SuppressWarnings("GroovyAccessibility")
+class ProtoNodeDecoratorSpec extends Specification {
+
+    ProtoNode decoratedProtoNode
+
+
+    def setup() {
```
sikarwarvimal: Here you have used `setup()` method that `run before every feature method` do you need to set before every feature method or this is just one time setup? 

Also note that `ProtoNodeDecorator can not be constructed with a null ProtoNode` feature is not using the `setup()` values but the `setup()` will fun for that feature too.

grabbit/src/test/groovy/com/twcable/grabbit/jcr/JCRNodeDecoratorSpec.groovy
```diff
@@ -0,0 +1,100 @@
...
+            getProperty(JCR_PRIMARYTYPE) >> Mock(Property) {
+                getString() >> { "nt:file" }
+            }
+        }
+
+        when:
+        final nodeDecorator = new JCRNodeDecorator(node)
+
+        then:
+        nodeDecorator.getPrimaryType() == "nt:file"
```
sikarwarvimal: i guess this could be `nodeDecorator.getPrimaryType() == JcrConstants.NT_FILE`

grabbit/src/test/groovy/com/twcable/grabbit/jcr/JCRNodeDecoratorSpec.groovy
```diff
@@ -0,0 +1,100 @@
...
+        then:
+        notThrown(RepositoryException)
+    }
+
+
+    def "getPrimaryType()"() {
+        given:
+        Node node = Mock(Node) {
+            getProperty(JCR_PRIMARYTYPE) >> Mock(Property) {
+                getString() >> { "nt:file" }
```
sikarwarvimal: `getString() >> { JcrConstants.NT_FILE }`

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoNodeDecoratorSpec.groovy
```diff
@@ -0,0 +1,127 @@
...
+        NodeProtos.Node.Builder nodeBuilder = NodeProtos.Node.newBuilder()
+        nodeBuilder.setName("somenode")
+
+        NodeProtos.Properties.Builder propertiesBuilder = NodeProtos.Properties.newBuilder()
+
+        NodeProtos.Property primaryTypeProperty = NodeProtos.Property
+                .newBuilder()
+                .setName(JCR_PRIMARYTYPE)
+                .setType(STRING)
+                .setValue(NodeProtos.Value.newBuilder().setStringValue("nt:unstructured"))
```
sikarwarvimal: `.setValue(NodeProtos.Value.newBuilder().setStringValue(JcrConstants.NT_UNSTRUCTURED))`

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoNodeDecoratorSpec.groovy
```diff
@@ -0,0 +1,127 @@
...
+        thrown(IllegalArgumentException)
+    }
+
+
+    def "Can get primary type"() {
+        when:
+        final protoNodeDecorator = new ProtoNodeDecorator(decoratedProtoNode)
+
+        then:
+        protoNodeDecorator.getPrimaryType() == "nt:unstructured"
```
sikarwarvimal: `protoNodeDecorator.getPrimaryType() == JcrConstants.NT_UNSTRUCTURED`

grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
jbornemann: This is just a defensive fail-fast gate, enforcing caller null-check responsibility.  All uses of this class are using the contract responsibly.  In the future, if this class is used per-contract, the Exception will never happen.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
jbornemann: All nodes have a primary type

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
jbornemann: All nodes have at least one property (jcr:primaryType), and per Jackrabbit, likely many more

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
jbornemann: Maybe.  I defer to @sagarsane as this is how it was before refactor.  Do users find this info useful?

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
jbornemann: That's a good question.  Protocol-wise we define it as possible for a property to not have a value.  JSR-283-wise, I'd have to dig.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
jbornemann: When this was first written, Sagar and I investigated a bit; but didn't find anything substantial.  Any ideas?

grabbit/src/test/groovy/com/twcable/grabbit/jcr/JCRNodeDecoratorSpec.groovy
jbornemann: Sounds like a good change

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoNodeDecoratorSpec.groovy
jbornemann: ``` decoratedProtoNode ``` is used in most of the features.  I'd like to use setup() to keep potential side-effects isolated.    Yeah, it's not used for "ProtoNodeDecorator can not be constructed with a null ProtoNode", but meh...the alternative doesn't read much better, and extra setup time I think is minimal.  Happy to change if you disagree.

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoNodeDecoratorSpec.groovy
jbornemann: Yeah.  I'll use JcrConstants

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy
jbornemann: Could you try again?  These lines don't correspond with places that could pop the stack

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy
jbornemann: Did you do something with setup() before you ran these tests?  Could you try again?

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoNodeDecoratorSpec.groovy
sikarwarvimal: As you are not mutating `decoratedProtoNode` for features so i guess one time setup would be sufficient by using `setupSpec() `. 

```
    @Shared
    ProtoNode decoratedProtoNode
    def setupSpec() {
    }
```

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy
sikarwarvimal: It was my local issue and now it is working fine.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
sikarwarvimal: Currently i do not have much exposure of these APIs but as i will work more I look into some solution.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
sikarwarvimal: ok

grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
sikarwarvimal: That's correct, it make sense.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
sikarwarvimal: I am trying to understand what will happen incase of the null (I guess edge cases)

Below class and method is indirectly using the `JCRNodeDecorator`. I do not see the exception has been handled anywhere during the call stack.
```<bean id="clientJcrNodesWriter" class="com.twcable.grabbit.client.batch.steps.jcrnodes.JcrNodesWriter" />```
```
    @Override
    void write(List<? extends ProtoNode> nodeProtos) throws Exception {
        Session session = theSession()
        for (ProtoNode nodeProto : nodeProtos) {
            writeToJcr(nodeProto, session)
        }
    }
```
So in this case Spring is executing this method, what will happen to the next node processing? I guess spring should handle the exception but i do not know for certain. Should we handle on our side too incase of exception?

File: grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy (hunk 6 above)
Comment (jdigger) [inline grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy]: Aesthetically, I prefer the `for` syntax here. Just feels clearer.

File: grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy (hunk 8 above)
Comment (jdigger) [inline grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy]: On the speed argument for `switch`: Java has [a special bytecode for `switch` (actually, two)](http://blog.jamesdbloom.com/JavaCodeToByteCode_PartOne.html#switch), so it does enjoy some advantages...

However, because Groovy lets you do a lot more with a `switch` statement than Java, it writes its byte code a bit differently. I haven't looked into what happens with `@CompileStatic`, which may give effectively identical bytecode for cases where there's no effectively difference between Groovy and Java.

All that said, though, the JIT is very, very smart. Especially for such a common construct. Similar to classic compiler techniques like loop-unrolling, it will convert `if else` sequences to the native equivalents to `switch` statements for you if doing so makes sense in context.

As always, go for the clearest way of communicating intent -- let the compiler worry about micro-efficiencies unless there's concrete evidence it's not able to do so in a specific case.

Comment (jdigger) [inline grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy]: In this specific situation, the comparison of a value to an enumerated list of potential values is conventionally handled with a `switch`, so I lean a little toward that here. If it were a single `if else` it would be easier/clearer to stay with the `if` structure. However, since this is `if else {if else}` it's inching into table-lookup territory.

File: grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy (hunk 12 above)
Comment (jdigger) [inline grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy]: I'll take a closer look at the logic a bit more later, but this is apropos to the discussion on `switch`: it's comparing to a list of potential values (not by equality, but can use Groovy's extension to use regex in a `case` statement), and it's missing the `default` case that would be more obvious if a `switch` were used...

File: grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy (hunk 2 above)
Comment (sagarsane) [inline grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy]: @sikarwarvimal we don't need to handle that. This is the Grabbit client side code. So, this method is never doing to get a list that is null. In fact, the `JcrNodesReader` that precedes this `JcrNodesWriter`, is *supposed* to return null when the inputStream is exhausted, but this Writer will never be called with a null list (and in-fact Spring documents this explicitly :-))

File: grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy (hunk 10 above)
Comment (sagarsane) [inline grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy]: Yeah... but I don't have a strong preference as such.

File: grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy (hunk 9 above)
Comment (sagarsane) [inline grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy]: These are protocol buffer compiler generated objects. They guarantee null safety (You can view them in `NodeProtos.java` in your IDE after you build your project)

File: grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy (hunk 7 above)
Comment (sagarsane) [inline grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy]: These are protocol buffer compiler generated objects. They guarantee null safety (You can view them in `NodeProtos.java` in your IDE after you build your project)

File: grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy (hunk 8 above)
Comment (jbornemann) [inline grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy]: switch it is!

File: grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy (hunk 13 above)
Comment (jbornemann) [inline grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy]: I'd rather duplicate in this case.  I think it reads more clear.

File: grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoNodeDecoratorSpec.groovy (hunk 18 above)
Comment (jbornemann) [inline grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoNodeDecoratorSpec.groovy]: I'd rather keep decoratedProtoNode isolated as to avoid setting future land mines for others writing additional features.

File: grabbit/src/test/groovy/com/twcable/grabbit/jcr/JCRNodeDecoratorSpec.groovy (hunk 19 above)
Comment (jbornemann) [inline grabbit/src/test/groovy/com/twcable/grabbit/jcr/JCRNodeDecoratorSpec.groovy]: Yeah, will change

File: grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy (hunk 12 above)
Comment (jbornemann) [inline grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy]:
```groovy
}
catch(ValueFormatException) {
    //Someone's being a pain
    final opinionatedProperty = node.getProperty(innerProtoProperty.name)
    //Are we being rebellious by trying to write a different type?  We should persist our changes?
    if(opinionatedProperty.type != innerProtoProperty.type) {
        //Away with it
        opinionatedProperty.remove()
        return this.writeToNode(node)    
    }
    //If we are trying to write to a property that accepts multiple values, lets succeed to its wishes
    if(opinionatedProperty.multiple) {

    }
    //Or act on the inverse
    else {

    }
}
```
might do it?  According to the documentation, these two cases are the only cases :

"if <code>value</code> cannot be converted to the specified type or if the property already exists and is multi-valued."

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
jbornemann: "10.4.2.4 No Null Values
Every property must have a value. The range of property states does not include having a “null value”, or “no value”. Setting a property to “null” is equivalent to removing that property (see §10.9 Removing Nodes and Properties)."

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
sagarsane: So, a bit of a background on why this was done. There were cases in the existing environments .. where some higher environments had the same property name with different types (so Name vs Name[] or String vs String[]) and when Grabbit ran, it was trying to write a property name with different property type. This should not really happen "normally"

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoNodeDecoratorSpec.groovy
jdigger: Agreed: `@Shared` should be avoided except when the setup time is unacceptably long.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
sagarsane: So, we are not dealing with JCR directly here. This is a "ProtoProperty". These are protocol buffer compiler generated objects. They guarantee null safety (You can view them in`NodeProtos.java` in your IDE after you build your project)

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
jdigger: May be worth adding a comment along those lines.

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoNodeDecoratorSpec.groovy
sikarwarvimal: I believe it is a good candidate for `@shared` because it does add to more unnecessary setup but i am fine otherwise too as the results remains the same.

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy
sikarwarvimal: I doubt `duplicate` + `hardcoded` = `clarity`

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
sikarwarvimal: Yes it seems protocol buffer is assigning ` value_ = java.util.Collections.emptyList();` for this so no NPE. Thanks for clarifying.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
sikarwarvimal: agree

grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
sikarwarvimal: Sounds good.

grabbit/src/test/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecoratorSpec.groovy
jdigger: I take the point of "I doubt `duplicate` + `hardcoded` = `clarity`", but in this specific case I agree with Jeff -- it's clearer/cleaner than the alternatives.

To take the use of `where:`, the "data-driven" approach makes things **_more_** confusing when used gratuitously, such as when there's only one data-set. It would be much better to simply declare them as variables/constants at the start of the method since they will only have one value, and it's much more idiomatic and shorter than declaring the `where:` section.

Even taking the simpler "declare a variable" approach, that adds more to the "mental stack" of things to keep track of, as well as adding the need to "dereference" the variable when reading it. Neither is a big deal, but it does have a small cost.

Now consider why "magic numbers" (of which hard coded duplicated values of any kind fall into the same category) are a bad thing: 1) you want a semantic clue what the value means, and 2) it's useful to have a common reference when the same logical thing is used in multiple places through the codebase (both to aid in changes, but also just to follow coherence).

For # 1 in the situation, the context and the value itself makes it more obvious than a name would. And for # 2 it's all scoped within a single (short) method.

I agree that duplicated hard-coded values are a pretty bad code-smell, but like all "smells" it doesn't mean it's _ipso facto_ wrong. Sometimes it's worth considering it compared to its alternatives and thinking more about what makes the pattern smelly. In **this particular case** it doesn't run into the problems with magic numbers, and the alternatives add further indirection with no clear benefit.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
jdigger: This is another place where a quick comment in the code would probably be helpful to get in front of future confusion.

jdigger:
Lots of discussion, but it looks like now there's just the `catch` to clean up, as well as some comments to add. :-)

PR #33, opened 2015-06-11
Bug fix in Client JcrNodesWriter when node name appears more than once in the path

sagarsane:
Found this bug where if the current node path is like `/foo/blah/bar/foo` .. and the child `/foo` is of type `nt:file`, we were effectively adding that child to `/blah/bar` instead of `/foo/blah/bar` 

Also added a test for this case.

@jdigger @jazeren1

### DIFF
--- a/grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
+++ b/grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
@@ -102,11 +102,13 @@ class JcrNodesWriter implements ItemWriter<NodeProtos.Node>, ItemWriteListener {
         log.debug "Primary Type: ${primaryType}"
 
         if (primaryType == NT_FILE) {
-            def temp = nodeProto.name.split("/")
-            final String fileName = temp.last()
-            final String parentName = nodeProto.name.replaceAll("/${fileName}", "")
-            final JcrNode parentNode = JcrUtils.getOrCreateByPath(parentName, null, session)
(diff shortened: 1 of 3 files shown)


grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
```diff
@@ -102,11 +102,11 @@ class JcrNodesWriter implements ItemWriter<NodeProtos.Node>, ItemWriteListener {
         log.debug "Primary Type: ${primaryType}"
 
         if (primaryType == NT_FILE) {
-            def temp = nodeProto.name.split("/")
-            final String fileName = temp.last()
-            final String parentName = nodeProto.name.replaceAll("/${fileName}", "")
```
sagarsane: This was the buggy logic.

sagarsane:
@jdigger @jazeren1 I think this PR is valid right now, but once #34 is handled and implemented, this code will probably completely go away / change as the `parent to child` node consideration will be implemented per JCR Specs.

jbornemann:
@sagarsane , @jdigger and I had some discussions this week, and feel this will still be valuable to v2.x users.  #34 will likely introduce a new major version, as client/server communication contract will likely change.

sagarsane:
Yeah, it will most likely change.

jbornemann:
This looks good.  Verified and tested.  Added some comments to make the negative indices list access syntax more clear to others.  Also added a quick commit that excludes fluent-hc from being pulled in to the project - It was being flagged by the vB task.

sagarsane:
Cool thanks for looking @jbornemann !

jdigger:
+1 CR

sagarsane:
Thanks Jim. Merging this in.

PR #15, opened 2015-03-10
 Trying with different travis configs

sagarsane:
Working to resolve #5 

@jazeren1 @jdigger

PR #37, opened 2015-06-29
Fix TravisCI build

jbornemann:
- Builds are passing when they shouldn't.  Probably due to running the Gradle daemon.
- We weren't installing protoc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
-language: java
+language: groovy
+
+addons:
+  apt:
+    packages:
+      - protobuf-compiler
 
 jdk:
   - oraclejdk7
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ addons:
     packages:
       - protobuf-compiler
 
+#Yeah, we are going to want to cache our build dependencies so our builds aren't held up downloading for every container
+cache:
+  directories:
+    - $HOME/.gradle
+
 jdk:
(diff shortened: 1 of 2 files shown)


jbornemann:
Trying to get the Gradle Daemon to stop spawning.  Travis must be doing something internally.

jdigger:
See #15 ...

jbornemann:
@jdigger @sagarsane take a look if you have the time.  This should solve our build woes.

jdigger:
With the "wrapper wrapper", do we still need to disable the daemon?

jdigger:
I suspect we can remove the disabling the daemon code since the wrapped wrapper should handle exit detection. Otherwise, looks great. Thanks!

jbornemann:
Yeah, I was being paranoid :)  I took it out of the wrapped wrapper.  

Would you prefer the commits to be merged as is, or squashed?

sagarsane:
I see that Travis CI is not happy with the last commit @jbornemann.

jbornemann:
I've seen "com.twcable.grabbit.client.batch.workflows.WorkflowManagerSpec > Should manage multithreading correctly" fail intermittently

jbornemann:
https://github.com/TWCable/grabbit/issues/38

sagarsane:
Okay.

sagarsane:
Changes look good though. +1

PR #47, opened 2015-07-20
Added missing copyright headers, and changed logic around ambiguous node property value types

jbornemann:
Changes : 

* Changed to not key off message in ValueFormatException.  Rather than "retrofitting" the property value coming across the wire, to the property value cardinality on the client (e.g ["foo", "bar"] -> "foo", or "foo" -> ["foo"]), we simply rewrite, and accept the property value cardinality type of the server.   

* Handles the case of a server sending a completely different property value type for a given property name.   

* Added missing copyright headers


jbornemann:
@jdigger @sagarsane @sikarwarvimal

sikarwarvimal:
Looks good, CR +1

Issue #20, opened 2015-03-27
Introduce a "deleteBeforeWrite" flag in Grabbit Configs


sagarsane:
This flag will be per path in a grabbit configuration. If this flag is enabled for a particular path, then `that path will be deleted from the client before writing`.

The configuration can look something like this : 
```json
{
	"path" : "/content/someContent",
	"workflowConfigIds" :
	    [
	        "/etc/workflow/launcher/config/update_asset_mod",
	        "/etc/workflow/launcher/config/update_asset_create",
	        "/etc/workflow/launcher/config/dam_xmp_nested_writeback",
	        "/etc/workflow/launcher/config/dam_xmp_writeback"
	    ],
	 "deleteBeforeWrite" : true
}
```

cc: @jazeren1 @jdigger

jazeren1:
@sagarsane can we assume that flag is optional?  Meaning if it's not there then we assume it's value is false?  (reason I ask this question is so that we don't have to update current filters to have this flag unless we want to.)

sagarsane:
Yeah that should be easy to do.

jdigger:
One thing to take into account: would #18 influence this? I don't think it would, but a destructive operation coming from the server is worth being aware of. (The overwrites are destructive too, which is why I don't think it's a big deal.) But when #17 is implemented maybe it should call-out that paths would be deleted?

sagarsane:
Yeah I agree. For #18, it should definitely be explicit as to "what all will happen?" with the pre-defined job sets. And yeah along with that, #17 should also call that out.

jdigger:
Just added #23 (exclude paths) which would also have implications on this...

PR #48, opened 2015-07-20
Delete before write

jbornemann:
Changes Made (Separate commits) : 

- Delete before write functionality
- Workflow_OFF/deleteBeforeWrite steps are executed concurrently (as no dependency on thread shared inputstream) 
- Updated README to reflect delete before write functionality 
- Renamed "PreprocessTasklet, Preprocess*, etc" to "NamespaceSyncTasklet, Namespace*, etc" as there are more than one preprocessors now, so the semantics became confusing.  
 
For https://github.com/TWCable/grabbit/issues/20

Please review @jdigger @sikarwarvimal @sagarsane

### DIFF
--- a/grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
+++ b/grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
@@ -1,0 +1,214 @@
+package com.twcable.grabbit.client.batch.steps.workspace
+
+import com.twcable.grabbit.jcr.JcrUtil
+import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+import org.apache.sling.jcr.api.SlingRepository
+import org.springframework.batch.core.StepContribution
+import org.springframework.batch.core.scope.context.ChunkContext
+import org.springframework.batch.core.step.tasklet.Tasklet
+import org.springframework.batch.repeat.RepeatStatus
+
+import javax.annotation.Nonnull
+import javax.annotation.Nullable
+import javax.jcr.Node
+import javax.jcr.PathNotFoundException
+import javax.jcr.Session
+
+/*
+ * Copyright 2015 Time Warner Cable, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * A client preSyncFlow tasklet that deletes nodes under the job path
+ * <p>
+ *     This tasklet is used to clear the workspace that will be written to by the job path.
+ *     If exclude paths are provided, it does not delete those paths, rather it deletes the nodes
+ *     around the exclude paths.  If no exclude paths are provided, the entire tree under the
+ *     job path is deleted.
+ *
+ *     This tasklet is activated by the "deleteBeforeWrite" path configuration
+ * </p>
+ */
+@Slf4j
+@CompileStatic
+class DeleteBeforeWriteTasklet implements Tasklet {
+
+    private String jobPath
+
+    private Collection<String> relativeExcludePaths
+
+    private SlingRepository slingRepository
+
+
+    /**
+     * @param jobPath the job path to evaluate.  Comes from the job parameters
+     * @param slingRepository comes from bean defined in client-osgi-config.xml
+     * @param excludePaths a '*' delimited string containing the paths to exclude.  Comes from the job parameters
+     */
+    DeleteBeforeWriteTasklet(@Nonnull final String jobPath, @Nonnull final SlingRepository slingRepository, @Nullable String excludePaths) {
+        this.slingRepository = slingRepository
+        //Ensures the job path leads with a '/'
+        this.jobPath = cleanJobPath(jobPath)
+        //Takes the * delimited paths string, creates a collection of relative paths, and normalizes
+        this.relativeExcludePaths = createExcludePaths(this.jobPath, excludePaths)
+    }
+
+
+    /**
+     * @param paths '*' delimited, expected to be driven from job parameters e.g /foo/bar/blah{@literal *}/foo/bar/meh
+     * @param jobPath the job path, used to make sure we are building well-formed relative paths e.g /foo/bar
+     * @return A collection of the exclude paths that are guaranteed to lead with a '/', e.g [blah, meh].  If null or empty, returns an empty collection
+     * @throws IllegalStateException if any paths are not well-formed to the jobPath
+     */
+    private static Collection<String> createExcludePaths(@Nonnull final String jobPath, @Nullable String paths) {
+        final thisPathsString = paths?.trim()
+        if(!thisPathsString) {
+            return []
+        }
+        final theExcludePaths = thisPathsString.split("\\*") as Collection<String>
+        final jobPathWithSlash = jobPath + '/'
+
+        if (theExcludePaths.any {!it.startsWith(jobPathWithSlash)}) throw new IllegalStateException("Not all exclude paths start with \"${jobPath}\": ${theExcludePaths}")
+
+        return theExcludePaths.collect { it.substring(jobPathWithSlash.length()) }
+    }
+
+    /**
+     * @param jobPath expected to be driven from job parameters
+     * @return a "cleaned" job path, i.e leads with a '/', no trailing '/', and trimmed, e.g "/foo/bar"
+     */
+    private static String cleanJobPath(String jobPath) {
+        String thisJobPath = jobPath.trim()
+        //Add the leading '/' if not present
+        thisJobPath = (thisJobPath[0] != '/') ? "/${thisJobPath}" : thisJobPath
+        //Remove any trailing '/'
+        thisJobPath = (thisJobPath[-1] != '/') ? thisJobPath : thisJobPath[0..-2] as String
+
+        return thisJobPath
+    }
+
+
+    /**
+     * Given the current context in the form of a step contribution, do whatever
+     * is necessary to process this unit inside a transaction. Implementations
+     * return {@link RepeatStatus#FINISHED} if finished. If not they return
+     * {@link RepeatStatus#CONTINUABLE}. On failure throws an exception.
+     *
+     * @param contribution mutable state to be passed back to update the current
+     * step execution
+     * @param chunkContext attributes shared between invocations but not between
+     * restarts
+     * @return an {@link RepeatStatus} indicating whether processing is
+     * continuable.
+     */
+    @Override
+    RepeatStatus execute(StepContribution contribution, ChunkContext chunkContext) throws Exception {
+        final Session session = JcrUtil.getSession(slingRepository, "admin")
+
+        final Node rootNode
+        try {
+            rootNode = session.getNode(jobPath)
+        }
+        catch(PathNotFoundException ex) {
+            log.warn "deleteBeforeWrite was enabled for jobPath ${jobPath} but job path does not exist on client!"
+            return RepeatStatus.FINISHED
+        }
+
+        //If we don't have any exclude paths, we don't have to worry about only clearing particular subtrees
+        if(!relativeExcludePaths || relativeExcludePaths.empty) {
+            rootNode.remove()
+        }
+        else {
+            deletePartialTree(rootNode, relativeExcludePaths)
+        }
+
+        session.save()
+        return RepeatStatus.FINISHED
+    }
+
+
+    /**
+     * Called recursively deleting any nodes under rootNode, that aren't apart of any relativeExcludePaths
+     */
+    private void deletePartialTree(@Nonnull final Node rootNode, @Nullable Collection<String> relativeExcludePaths) {
+        //Base case.  If a root node does not have any exclude paths under it, it must have been excluded
+        if(!relativeExcludePaths) {
+            return
+        }
+        //Compute the current tree level excluded nodes, and compute the remaining paths to traverse
+        Collection<NodeAndExcludePaths> nodeAndExcludePaths = relativeExcludePaths.inject([] as Collection<NodeAndExcludePaths>) { def acc, def thisPath ->
+            final thisNodeAndExcludePath = NodeAndExcludePaths.fromPath(thisPath)
+            //If we have already created a nodeAndExcludePath for this node name, just add to its exclude paths
+            final matchingNodeName = acc.find { it.nodeName == thisNodeAndExcludePath.nodeName }
+            if(matchingNodeName) {
+                matchingNodeName.excludePaths.addAll(thisNodeAndExcludePath.excludePaths)
+            }
+            else {
+                acc.add(thisNodeAndExcludePath)
+            }
+            return acc
+        } as Collection
+        //Delete nodes allowed under this tree
+        final rootNodeChildren = rootNode.getNodes()
+        while(rootNodeChildren.hasNext()) {
+            final currentNode = rootNodeChildren.nextNode()
+            //If this node is in our exclusion list, don't delete it
+            if(!(nodeAndExcludePaths.find { currentNode.name == it.nodeName })) {
+                currentNode.remove()
+            }
+        }
+        //Recurse on each
+        nodeAndExcludePaths.each {
+            try {
+                deletePartialTree(rootNode.getNode(it.nodeName), it.excludePaths)
+            }
+            catch(PathNotFoundException ex) {
+                log.warn "Exclude node ${it.nodeName} of parent ${rootNode.path} does not exist on client, so we won't traverse it for adjacent node deletion"
+            }
+        }
+    }
+
+    /**
+     * If we have a path for a node such as 'foo/bar/doo'
+     * then node is 'foo', and the relativeExcludePaths is 'bar/doo'.  As other relativeExcludePaths are inspected, we may find that
+     * there are NodeAndExcludePaths with the same nodeName, in which case they will be merged by adding to an instance's excludePaths
+     */
+    static class NodeAndExcludePaths {
+
+        String nodeName
+
+        //Can mutate in the case described above
+        Collection<String> excludePaths
+
+        /**
+         * @param path Expects a relative path e.g 'foo/bar/doo'
+         * @return NodeExcludePath(nodeName: "foo", excludePaths: "bar/doo")
+         */
+        static NodeAndExcludePaths fromPath(final String path) {
+            final String thisPath = path
+            //We need to find the current node, and the remaining path e.g 'foo/bar/doo', node: foo, remaining path: /bar/doo
+            final slashIndex = thisPath.indexOf('/')
+            //There is no remaining path
+            if(slashIndex == -1) {
+                return new NodeAndExcludePaths(nodeName: thisPath, excludePaths: [])
+            }
+            final remainingPath = thisPath.substring(slashIndex+1)
+            final node = thisPath.substring(0, slashIndex)
+            return new NodeAndExcludePaths(nodeName: node, excludePaths: [remainingPath])
+        }
+    }
+
+}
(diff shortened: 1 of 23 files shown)


sikarwarvimal:
@jbornemann is this ready for review and testing?

sikarwarvimal:
It is a old branch but it seems you are still working on this?

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@Slf4j
+@CompileStatic
+class DeleteBeforeWriteTasklet implements Tasklet {
```
jdigger: Can we get a javadoc for the class?

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+
+    private String jobPath
+
+    private Collection<String> excludePaths
+
+    private SlingRepository slingRepository
+
+    void setExcludePaths(String paths) {
```
jdigger: Should add javadocs explaining the format that this is expected to be in, referencing the fact that `ClientBatchJob` is what populates the JobParameters with this as a single string.

jdigger: Given that I am anal when it comes to immutability and such: This Tasklet is meant to be immutable, so these values should all be passed into the constructor, not setters.
```xml
<constructor-arg name="jobPath" value="#{jobParameters[path]}"/>
```

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+        }
+        //Recurse on each
+        nodeAndExcludePaths.each {
+            deletePartialTree(rootNode.getNode(it.nodeName), it.excludePaths)
+        }
+    }
+
+    /**
+     * If we have a path for a node such as 'foo/bar/doo'
+     * node is 'foo', and the excludePaths is '/bar/doo'
```
jdigger: There should be a "then" joining those two lines.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+     */
+    static class NodeAndExcludePaths {
+
+        String nodeName
+        Collection<String> excludePaths
+
+        static NodeAndExcludePaths fromPath(final String path) {
+            //Remove leading '/'
+            final String thisPath = path.replaceFirst('/', '')
```
jdigger: Actually, given the example in the javadoc, this would mess things up.

"foo/bar/doo" -> {"foobar", "/doo"}

jdigger: Safer:
```groovy
final thisPath = path.startsWith('/') ? path.substring(1) : path
```

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * If we have a path for a node such as 'foo/bar/doo'
+     * node is 'foo', and the excludePaths is '/bar/doo'
+     */
+    static class NodeAndExcludePaths {
+
+        String nodeName
+        Collection<String> excludePaths
```
jdigger: Either make this immutable (which I see from line 112 it is not) or explicitly declare that it's expected to mutate.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+    void setExcludePaths(String paths) {
+        final thisPathsString = paths?.trim()
+        if(!thisPathsString) {
+            excludePaths = []
+            return
+        }
+        Collection<String> theseExcludePaths = thisPathsString.split("\\*") as Collection
+        excludePaths = theseExcludePaths.collect { String excludePath ->
+            return (excludePath[0] != '/') ? "/${excludePath}".toString() : excludePath
+        }
```
jdigger: This is better replaced by
```groovy
def jobPathSlash = jobPath + '/'
if (excludePaths.any {!it.startsWith(jobPathSlash)})
    throw new IllegalArgumentException("Not all exclude paths start with \"${jobPathSlash}\": ${excludePaths}")
```
because then (assuming things have been changed over to constructor form and the valid format of `jobPath` has been assured) the real purpose of this check (that the exclude paths are all absolute) will be met, as well as guaranteeing that they are all relative to the jobPath.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+
+    private void deletePartialTree(Node rootNode, Collection<String> relativeExcludePaths) {
+        //Base case.  If a root node does not have any exclude paths under it, it must have been excluded
+        if(!relativeExcludePaths) {
+            return
+        }
+        //Compute the current tree level excluded nodes, and compute the remaining paths to traverse
+        Collection<NodeAndExcludePaths> nodeAndExcludePaths = relativeExcludePaths.inject([] as Collection<NodeAndExcludePaths>) { def acc, def thisPath ->
+            final thisNodeAndExcludePath = NodeAndExcludePaths.fromPath(thisPath)
+            //If we have already created a nodeAndExcludePath for this node name, just add to it's exclude paths
```
jdigger: Grammar Nazi: "just add it to it is paths?" :speak_no_evil:

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
jdigger: Even better, as I continue to think through this, would be
```groovy
def theExcludePaths = thisPathsString.split("\\*") as Collection<String>
def jobPathSlash = jobPath + '/'
if (theExcludePaths.any {!it.startsWith(jobPathSlash)}) throw new IllegalStateException("Not all exclude paths start with \"${jobPath}\": ${theExcludePaths}")
this.excludePaths = theExcludePaths.collect { it.substring(jobPathSlash.length()) }
```
then you can always treat the exclude paths as relative (since later they are converted that way anyway) **_and_** you've already validated that they are legit for the job path.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+        else {
+            deletePartialTree(session)
+        }
+        session.save()
+        return RepeatStatus.FINISHED
+    }
+
+
+    private void deletePartialTree(final Session session) {
+        final relativeExcludePaths = excludePaths.collect { it - jobPath }
```
jdigger: This is no longer needed if `excludePaths` is set to be relative when set.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+            //Remove leading '/'
+            final String thisPath = path.replaceFirst('/', '')
+            //We need to find the current node, and the remaining path e.g 'foo/bar/doo', node: foo, remaining path: /bar/doo
+            final slashIndex = thisPath.indexOf('/')
+            //There is no remaining path
+            if(slashIndex == -1) {
+                return new NodeAndExcludePaths(nodeName: thisPath, excludePaths: [])
+            }
+            final remainingPath = thisPath.substring(slashIndex)
+            final node = thisPath - remainingPath
```
jdigger: More efficient than the `minus(..)`
```groovy
final node = thisPath.substring(0, slashIndex)
```
also a bit clearer as to the intent and less likely to get caught on subtle bugs like `remainingPath` happening to be a substring in what is supposed to be `node`

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+        static NodeAndExcludePaths fromPath(final String path) {
+            //Remove leading '/'
+            final String thisPath = path.replaceFirst('/', '')
+            //We need to find the current node, and the remaining path e.g 'foo/bar/doo', node: foo, remaining path: /bar/doo
+            final slashIndex = thisPath.indexOf('/')
+            //There is no remaining path
+            if(slashIndex == -1) {
+                return new NodeAndExcludePaths(nodeName: thisPath, excludePaths: [])
+            }
+            final remainingPath = thisPath.substring(slashIndex)
```
jdigger: If you add 1 to the index then you're guaranteed to get a true relative path for the remaining path
```groovy
final remainingPath = thisPath.substring(slashIndex+1)
```

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
jdigger: If `remainingPath` below, and `excludePaths` above are both guaranteed to be relative, then this whole piece about removing a leading `/` is no longer needed.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,158 @@
...
+        Collection<String> theseExcludePaths = thisPathsString.split("\\*") as Collection
+        excludePaths = theseExcludePaths.collect { String excludePath ->
+            return (excludePath[0] != '/') ? "/${excludePath}".toString() : excludePath
+        }
+    }
+
+
+    void setJobPath(String jobPath) {
+        String thisJobPath = jobPath.trim()
+        thisJobPath = thisJobPath[0] != '/' ? "/${thisJobPath}" : thisJobPath
```
jdigger: May want to add
```groovy
thisJobPath = (thisJobPath[-1] != '/') ? thisJobPath : thisJobPath[0..-2]
```
to guard against `/foo/` being passed in.

jdigger:
What should happen if a job path of `/fooble` is passed in but there's no path in the JCR by that name?

What should happen if a job path of `/foo` is passed, which exists, but an exclude path of `/foo/bar` is given but there's no path in the JCR by that name?

jdigger:
FWIW, I love the use of comments in the various sections...

jbornemann:
@jdigger ready for another round

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,214 @@
...
+     * @param contribution mutable state to be passed back to update the current
+     * step execution
+     * @param chunkContext attributes shared between invocations but not between
+     * restarts
+     * @return an {@link RepeatStatus} indicating whether processing is
+     * continuable.
+     */
+    @Override
+    RepeatStatus execute(StepContribution contribution, ChunkContext chunkContext) throws Exception {
+        final Session session = JcrUtil.getSession(slingRepository, "admin")
```
sagarsane: Do we need a new session here? Can we not use the one persisted in `ThreadLocal`? (ClientBatchJobContext) ?

grabbit/src/main/resources/META-INF/spring/client-batch-job.xml
```diff
@@ -32,16 +32,35 @@
...
     <batch:job id="clientJob" xmlns="http://www.springframework.org/schema/batch" job-repository="clientJobRepository">
-        <batch:step id="clientPreprocessor" next="clientWorkflowOff">
-            <batch:tasklet ref="clientPreprocessTasklet" transaction-manager="clientTransactionManager"/>
-        </batch:step>
-        <batch:step id="clientWorkflowOff" next="clientJcrNodes">
-            <batch:tasklet ref="clientWorkflowOffTasklet" transaction-manager="clientTransactionManager"/>
+        <!--  Do any presync tasks concurrently that we can, e.g turning off workflows -->
+        <batch:split id="preSyncFlow" task-executor="asyncTaskExecutor" next="clientNamespaceSync">
+            <batch:flow>
+                <batch:step id="clientWorkflowOff">
```
sagarsane: `clientWofkflowOff` is VERY lightweight step. Having said that, even if it is light weight, it actually does manage concurrency internally (so there is already multi-threading handling involved in the actual step and on that we are spawning different thread for the step it self).

Do we know this split of steps is going to give us tangible benefits?

If not, then I vote we keep it simple and have the steps be sequential. If there are no tangible benefits, I think this way will be harder to maintain (as you know we already have 1 Thread Per Path in the config)

cc: @jdigger

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,214 @@
...
+            deletePartialTree(rootNode, relativeExcludePaths)
+        }
+
+        session.save()
+        return RepeatStatus.FINISHED
+    }
+
+
+    /**
+     * Called recursively taking a rootNode, and deleting any nodes adjacent except adjacent corresponding nodes form relativeExcludePaths
```
sagarsane: I feel `and deleting any nodes adjacent except adjacent corresponding nodes form relativeExcludePaths` is a bit confusing and could be clearer :-)

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,214 @@
...
+
+        //If we don't have any exclude paths, we don't have to worry about only clearing particular subtrees
+        if(!relativeExcludePaths || relativeExcludePaths.empty) {
+            rootNode.remove()
+        }
+        else {
+            deletePartialTree(rootNode, relativeExcludePaths)
+        }
+
+        session.save()
```
sagarsane: This is just a suggestion (and open for discussion too :) )but you could implement a `DeleteBeforeWriteStepExecutionListener` that prepares (in this case, I guess there's nothing much to prepare; but you can do anything in it) and cleans up after the step (in this case would be `session.save()`)

May be it will be cleaner? :) (example : https://github.com/TWCable/grabbit/blob/master/grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesStepExecutionListener.groovy )

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
```diff
@@ -0,0 +1,214 @@
...
+            rootNode = session.getNode(jobPath)
+        }
+        catch(PathNotFoundException ex) {
+            log.warn "deleteBeforeWrite was enabled for jobPath ${jobPath} but job path does not exist on client!"
+            return RepeatStatus.FINISHED
+        }
+
+        //If we don't have any exclude paths, we don't have to worry about only clearing particular subtrees
+        if(!relativeExcludePaths || relativeExcludePaths.empty) {
+            rootNode.remove()
```
sagarsane: Wondering if there is a preference / difference between `node.remove()` and `session.removeItem(/path)` ?

grabbit/src/main/resources/META-INF/spring/client-batch-job.xml
jbornemann: Performance benefits are marginal at this point; but any small wins are great, especially since the complexity is managed by Spring Batch.  That said it's more to establish the pattern, that we should be doing independent tasks concurrently.

I hope to move namespaceSync into preFlowSync eventually, but for the sake of keeping the PR small...This will make this change a bit more significant.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
jbornemann: Oops, typo.  

Open to suggestions :-). I'll try to make it more clear

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/workspace/DeleteBeforeWriteTasklet.groovy
jbornemann: This isn't managed by SimpleSynchronousTaskExecutioner (default) like clientJcrNodes, and namespaceSync are.  It's managed by it's async counterpart, so we can't rely on our ThreadLocal "hack".  

The session really only needs to be shared in clientJcrNodes in my opinion.  You probably know best with your research, but wondering if there is a better way to manage a session within clientJcrNodes ( Setting as a task executioner property maybe?) It's sort of an unfortunate leaky abstraction we have to rely on

sagarsane: Yes I know using ThreadLocal is kind of a last resort general but there were several things that made us use ThreadLocal instead of anything else.

1. The design was simpler to have 1 session per path and we had to share the incoming input stream between steps anyways (not just `clientJcrNodes`)
2. Originally, I was trying to use Spring Batch's `ExecutionContext` to store the session (and the input stream). But, Spring Batch discourages placing large objects in its ExecutionContext, which is understandable since ExecutionContext is one of the objects persisted in the `Spring Batch's Job Repository`.. which for our case, we have our `JcrJobRepository` layer.
3. Hence, ThreadLocal was the last resort and if done properly, we know that it would always work and as mentioned above, we had to share the `inputStream` amongst the steps anyways.

grabbit/src/main/resources/META-INF/spring/client-batch-job.xml
sagarsane: I do think that using the feature will make sense in some cases, but I still don't think that parallelizing `clientWorkflowOffTasklet` and `deleteBeforeWriteTasklet` makes sense (and may be its not safe either). 

For example : I think its perfectly reasonable to have a workflow that will trigger when a DAM Asset is removed (may be to remove `custom renditions` for that Asset or something).

For such a scenario, if we parallellize the two tasks above, it is possible that the nodes are removed before the corresponding workflows are turned off, which will result in undesirable effects.

I could however see `nameSpaceSync` and `clientWorkflowOff` tasks being parallelized. I think that will be safe, but the current structure I do not think is safe in all the cases.

What do others think @jdigger @jazeren1 @sikarwarvimal Paul?

jbornemann: Although unlikely, I could see a case for that.  I'll revert it back to synchronous, and we can explore making nameSpaceSync, and deleteBeforeWrite concurrent at a later date.

jbornemann: https://github.com/TWCable/grabbit/issues/54

sagarsane: Cool. The revert will be done in a separate PR I presume?

Issue #34, opened 2015-06-15
Handle all nt:hierarchy nodes properly


sagarsane:
Currently, `nt:file` is the only node that is explicitly handled (and that too incompletely). Grabbit currently only supports `nt:file` with underlying `jcr:content` of `nt:resource` type. According to the JCR specification, the underlying `jcr:content` node can be of ANY type.

Also, there are other hierarchy nodes that need to be handled / implemented.

http://wiki.apache.org/jackrabbit/nt%3AhierarchyNode

jdigger:
A couple more potentially useful references:

* https://github.com/apache/jackrabbit-filevault/blob/trunk/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/api/Aggregator.java
* https://docs.adobe.com/docs/en/cq/5-6-1/javadoc/com/day/cq/replication/DefaultAggregateHandler.html

Will likely need to change the top-level of the stream from being Nodes to being "NodeAggregate" or the like.

jdigger:
![nodes](https://cloud.githubusercontent.com/assets/3276/8684949/9332a032-2a38-11e5-8203-5b017b11d239.png)

```
message Node {
    required string name = 1;
    required Properties properties = 2;
    repeated Node children = 3; // new field
}
```

Given the above data structure (where H in an nt:hierarchy, C is a required child node, and N is a non-nt:hierarchy node), and the above tweek to [node.proto](../blob/master/grabbit/src/main/proto/node.proto), what would show on the protobuf stream (where `N` is short for Node) would be:

N(n:H1, c:[N(n:C1)]),
N(n:N1),
N(n:H2, c:[N(n:C2)]),
N(n:N2),
N(n:H3, c:[N(n:C3)])

The algorithm is depth-first traversal, where if a node is n:h then scan its children for any that return `child.nodeDef.isMandatory == true` and include them in the aggregate. Continue the traversal, but if `node.nodeDef.isMandatory == true` then assume it's already been handled by the parent so skip it.

[...]

PR #49, opened 2015-07-21
NT Hierarchy node logic including required nodes with new Node structure

sikarwarvimal:
Logic to process nt:hierarchy nodes with required child node has beed added. To process the given scenario `Node` proto structure has been modified.

This PR is ready for review and testing.

Test Scenario:
Identify any hierarchal node type like `nt:file` and try to sync that node and the sync will process the `nt:file` node and the required child node `jcr:content` as one transaction.

@jdigger @jbornemann @sagarsane @masroormohammed

### DIFF
--- a/grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
+++ b/grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
@@ -22,19 +22,11 @@ import com.twcable.grabbit.jcr.ProtoNodeDecorator
 import com.twcable.grabbit.proto.NodeProtos.Node as ProtoNode
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
-import org.apache.jackrabbit.commons.JcrUtils
-import org.apache.jackrabbit.value.BinaryValue
 import org.springframework.batch.core.ItemWriteListener
 import org.springframework.batch.item.ItemWriter
 import org.springframework.util.StopWatch
 
-import javax.jcr.Node
 import javax.jcr.Session
-import javax.jcr.nodetype.NodeType
-
-import static javax.jcr.Node.JCR_CONTENT
-import static org.apache.jackrabbit.JcrConstants.JCR_DATA
-import static org.apache.jackrabbit.JcrConstants.NT_FILE
 
 /**
  * A Custom ItemWriter that will write the provided Jcr Nodes to the {@link JcrNodesWriter#theSession()}
@@ -68,18 +60,17 @@ class JcrNodesWriter implements ItemWriter<ProtoNode>, ItemWriteListener {
     }
 
 
-    @Override
-    /*
+    /**
      *  The JcrNodesReader that funnels proto nodes into here
-      * will return null when the stream is finished to indicate completion, but Spring will pass null to this point
+     * will return null when the stream is finished to indicate completion, but Spring will pass null to this point
      */
+    @Override
     void write(List<? extends ProtoNode> nodeProtos) throws Exception {
         Session session = theSession()
         for (ProtoNode nodeProto : nodeProtos) {
             writeToJcr(nodeProto, session)
         }
     }
-
 
     private static <T> T withStopWatch(String stopWatchId, Closure<T> cl) {
         StopWatch stopWatch = new StopWatch(stopWatchId)
@@ -93,27 +84,16 @@ class JcrNodesWriter implements ItemWriter<ProtoNode>, ItemWriteListener {
         return retVal
     }
 
-
     private static void writeToJcr(ProtoNode nodeProto, Session session) {
-
         JCRNodeDecorator jcrNode = new ProtoNodeDecorator(nodeProto).writeToJcr(session)
-
-        //If the primary type is NT_FILE, we have to do some goofiness to make sure a jcr:content node is written with the nt_file node
-        if (jcrNode.getPrimaryType() == NT_FILE) {
-            //Now we have to add the jcr:content node to enforce the nt:hierarchy requirements
-            JCRNodeDecorator jcrContentNode = new JCRNodeDecorator(JcrUtils.getOrAddNode(jcrNode as Node, JCR_CONTENT, NodeType.NT_RESOURCE))
-            /*
-            * TODO : This is a workaround for the case where a chunk gets 'saved' in JCR and the last node was 'nt:file'
-            * If jcr:data is not part of that chunk then you will get a constraint violation exception
-            * To get around that, just add an empty binary jcr:data here with a "temp" value
-            * This will always be overridden by the actual jcr:data value as that will be the next thing received
-            */
-            jcrContentNode.setProperty(JCR_DATA, new BinaryValue("temp".bytes))
-            jcrContentNode.setLastModified()
-        }
         jcrNode.setLastModified()
+        // This will processed all mandatory child nodes only
+        if(nodeProto.mandatoryChildNodeList && nodeProto.mandatoryChildNodeList.size() > 0) {
+            for(ProtoNode childNode: nodeProto.mandatoryChildNodeList) {
+                writeToJcr(childNode, session)
+            }
+        }
     }
-
 
     private Session theSession() {
         ClientBatchJobContext clientBatchJobContext = ClientBatchJobContext.THREAD_LOCAL.get()
--- a/grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
+++ b/grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
@@ -25,11 +25,11 @@ import com.twcable.grabbit.proto.NodeProtos.Property
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
 import org.springframework.batch.item.ItemProcessor
-
 import javax.jcr.Node as JcrNode
 import javax.jcr.Property as JcrProperty
 import javax.jcr.PropertyType
 import javax.jcr.Value
+import javax.jcr.nodetype.ItemDefinition
 
 import static javax.jcr.PropertyType.BINARY
 import static javax.jcr.PropertyType.BOOLEAN
@@ -55,7 +55,6 @@ import static org.apache.jackrabbit.JcrConstants.JCR_PRIMARYTYPE
 class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
 
     private String contentAfterDate
-
 
     void setContentAfterDate(String contentAfterDate) {
         this.contentAfterDate = contentAfterDate
@@ -88,10 +87,65 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
             }
         }
 
-        final List<JcrProperty> properties = jcrNode.properties.toList()
-        Node.Builder nodeBuilder = Node.newBuilder()
-        nodeBuilder.setName(jcrNode.path)
+        // Skip this node because it has already been processed by its parent
+        if(isRequiredNode(jcrNode)) {
+            return null;
+        } else {
+            // Build parent node
+            return buildNode(jcrNode, getRequiredChildNodes(jcrNode))
+        }
+    }
 
+    /**
+     * Build node and "only" mandatory child nodes
+     * @param JCR Node
+     * @param Mandatory childNodes
+     * @return Node Proto node
+     */
+    private Node buildNode(JcrNode jcrNode, Collection<JcrNode> childNodes) {
+        final Node.Builder nodeBuilder = Node.newBuilder()
+        nodeBuilder.setName(jcrNode.path)
+        nodeBuilder.setProperties(buildProperties(jcrNode))
+        childNodes?.each {
+            nodeBuilder.addMandatoryChildNode(buildNode(it, getRequiredChildNodes(it)))
+        }
+        return nodeBuilder.build()
+    }
+
+    /**
+     * Identify all required child nodes
+     * @param JCR Node
+     * @return list of required child nodes
+     */
+    private static Collection<JcrNode> getRequiredChildNodes(JcrNode jcrNode) {
+        return hasMandatoryChildNodes(jcrNode) ? jcrNode.getNodes().findAll{JcrNode childJcrNode -> isRequiredNode(childJcrNode)} : null
+    }
+
+    /**
+     * Checks for mandatory child nodes
+     * @param JCR Node
+     * @return boolean flag for mandatory child node
+     */
+    private static boolean hasMandatoryChildNodes(JcrNode jcrNode) {
+        return jcrNode.primaryNodeType.childNodeDefinitions.any { ((ItemDefinition)it).mandatory }
+    }
+
+    /**
+     * Checks required node
+     * @param JCR node
+     * @return boolean flag for mandatory node
+     */
+    private static boolean isRequiredNode(JcrNode node) {
+        return node.definition.isMandatory()
+    }
+
+    /**
+     * Build Proto Property object from JCR node
+     * @param jcrNode for Proto properties
+     * @return Proto Property object
+     */
+    private static NodeProtos.Properties buildProperties(JcrNode jcrNode) {
+        final List<JcrProperty> properties = jcrNode.properties.toList()
         Properties.Builder propertiesBuilder = Properties.newBuilder()
         properties.each { JcrProperty jcrProperty ->
             //Before adding a property to the Current Node Proto message, check if the property
@@ -101,11 +155,8 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
                 propertiesBuilder.addProperty(property)
             }
         }
-        nodeBuilder.setProperties(propertiesBuilder.build())
-
-        nodeBuilder.build()
+        propertiesBuilder.build()
     }
-
     /**
      * Returns the "jcr:created", "jcr:lastModified" or "cq:lastModified" date property
      * for current Jcr Node
--- a/grabbit/src/main/proto/node.proto
+++ b/grabbit/src/main/proto/node.proto
@@ -9,6 +9,7 @@ option java_outer_classname = "NodeProtos";
 message Node {
     required string name = 1;
     required Properties properties = 2;
+    repeated Node mandatoryChildNode = 3;
 }
 
 message Properties {
--- a/grabbit/src/test/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriterSpec.groovy
+++ b/grabbit/src/test/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriterSpec.groovy
@@ -19,6 +19,7 @@ package com.twcable.grabbit.client.batch.steps.jcrnodes
 import com.twcable.grabbit.client.batch.ClientBatchJobContext
 import com.twcable.grabbit.jcr.JcrUtil
 import com.twcable.grabbit.proto.NodeProtos
+import org.apache.jackrabbit.JcrConstants
 import spock.lang.Specification
 import spock.lang.Subject
 
@@ -36,25 +37,44 @@ class JcrNodesWriterSpec extends Specification {
 
     def "Can get a Jcr File Node given a single Protobuf Message Node"() {
         given:
+
         NodeProtos.Node.Builder nodeBuilder = NodeProtos.Node.newBuilder()
         nodeBuilder.setName("/default.groovy")
         NodeProtos.Properties.Builder propertiesBuilder = NodeProtos.Properties.newBuilder()
         NodeProtos.Property aProperty = NodeProtos.Property
             .newBuilder()
-            .setName("jcr:primaryType")
+            .setName(JcrConstants.JCR_PRIMARYTYPE)
             .setType(STRING)
-            .setValue(NodeProtos.Value.newBuilder().setStringValue("nt:file"))
+            .setValue(NodeProtos.Value.newBuilder().setStringValue(JcrConstants.NT_FILE))
             .build()
 
         propertiesBuilder.addProperty(aProperty)
         aProperty = NodeProtos.Property
             .newBuilder()
-            .setName("jcr:lastModified")
+            .setName(JcrConstants.JCR_LASTMODIFIED)
             .setType(STRING)
             .setValue(NodeProtos.Value.newBuilder().setStringValue("Date"))
             .build()
         propertiesBuilder.addProperty(aProperty)
         nodeBuilder.setProperties(propertiesBuilder.build())
+
+        // Child Node jcr:content
+        NodeProtos.Node.Builder childNodeBuilder = NodeProtos.Node.newBuilder()
+        childNodeBuilder.setName("/default.groovy/$JcrConstants.JCR_CONTENT")
+        NodeProtos.Properties.Builder childPropertiesBuilder = NodeProtos.Properties.newBuilder()
+        NodeProtos.Property childProperty = NodeProtos.Property
+                .newBuilder()
+                .setName(JcrConstants.JCR_PRIMARYTYPE)
+                .setType(STRING)
+                .setValue(NodeProtos.Value.newBuilder().setStringValue(JcrConstants.NT_RESOURCE))
+                .build()
+        childPropertiesBuilder.addProperty(childProperty)
+
+        childNodeBuilder.setProperties(childPropertiesBuilder.build())
+
+        // Adding Child Node
+        nodeBuilder.addMandatoryChildNode(childNodeBuilder.build())
+
         NodeProtos.Node nodeProto = nodeBuilder.build()
 
         Session session = JcrUtil.getSession(repository().build(), "admin")
@@ -67,8 +87,8 @@ class JcrNodesWriterSpec extends Specification {
         JcrNode jcrNode = session.getNode("/default.groovy")
         jcrNode != null
         jcrNode.hasProperties()
-        JcrNode jcrAnotherNode = session.getNode("/default.groovy/${JcrNode.JCR_CONTENT}")
-        jcrAnotherNode != null
+        JcrNode childNode = session.getNode("/default.groovy/$JcrConstants.JCR_CONTENT")
+        childNode != null
 
     }
 
--- a/grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
+++ b/grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
@@ -18,14 +18,20 @@ package com.twcable.grabbit.server.batch.steps.jcrnodes
 
 import com.twcable.grabbit.proto.NodeProtos
 import com.twcable.jackalope.NodeBuilder as FakeNodeBuilder
+import com.twcable.jackalope.impl.jcr.ValueImpl
+import org.apache.jackrabbit.JcrConstants
+import org.apache.jackrabbit.commons.iterator.NodeIteratorAdapter
+import org.apache.jackrabbit.commons.iterator.PropertyIteratorAdapter
 import spock.lang.Specification
 import spock.lang.Subject
-
 import javax.jcr.Node as JcrNode
+import javax.jcr.Property as JcrProperty
+import javax.jcr.PropertyIterator
+import javax.jcr.nodetype.NodeDefinition
+import javax.jcr.nodetype.NodeType
 
 import static com.twcable.jackalope.JCRBuilder.node
 import static com.twcable.jackalope.JCRBuilder.property
-import static com.twcable.jackalope.JcrConstants.NT_FILE
 import static javax.jcr.PropertyType.LONG
 import static javax.jcr.PropertyType.STRING
 
@@ -35,16 +41,16 @@ class JcrNodesProcessorSpec extends Specification {
     def "Can marshall a JCR Node to a Protobuf Message"() {
         given:
         FakeNodeBuilder fakeNodeBuilder =
-            node("default.groovy",
-                node("jcr:content",
-                    property("jcr:data", "foo")
-                ),
-                property("jcr:primaryType", NT_FILE),
-                property("jcr:lastModified", "Date"),
-                property("multiValueLong", [1L, 2L, 4L] as Object[]),
-                property("multiValueString", ["a", "b", "c"] as Object[]),
+                node("default.groovy",
+                        node("jcr:content",
+                                property("jcr:data", "foo")
+                        ),
+                        property("jcr:primaryType", JcrConstants.NT_FILE),
+                        property("jcr:lastModified", "Date"),
+                        property("multiValueLong", [1L, 2L, 4L] as Object[]),
+                        property("multiValueString", ["a", "b", "c"] as Object[]),
 
-            )
+                )
         JcrNode aJcrNode = fakeNodeBuilder.build()
 
         when:
@@ -69,4 +75,120 @@ class JcrNodesProcessorSpec extends Specification {
         propertyString.type == STRING
     }
 
+    def "Can marshall a JCR Node nt:file with mandatory child node nt:resource to a Protobuf Message "() {
+        given:
+        def imageFile = "/content/geometrixx-outdoors/jcr:content/image/file/jcr:content/dam:thumbnails/dam:thumbnail_600.png"
+        def parentNode = createNode(imageFile, false, JcrConstants.NT_FILE,
+                [createNode("${imageFile}/jcr:content", true, JcrConstants.NT_RESOURCE)])
+
+        when:
+        NodeProtos.Node nodeProto = new JcrNodesProcessor().process(parentNode)
+
+        then:
+        nodeProto.name == imageFile
+        nodeProto.properties.propertyList.first().value.stringValue == JcrConstants.NT_FILE
+        nodeProto.mandatoryChildNodeList.size() == 1
+        nodeProto.mandatoryChildNodeList.first().name == "${imageFile}/jcr:content"
+        nodeProto.mandatoryChildNodeList.first().properties.propertyList.first().name == JcrConstants.JCR_PRIMARYTYPE
+        nodeProto.mandatoryChildNodeList.first().properties.propertyList.first().value.stringValue == JcrConstants.NT_RESOURCE
+    }
+
+    def "Can marshall a JCR Node nt:file with mandatory child node nt:unstructured to a Protobuf Message "() {
+        given:
+        def thumbnailFile = "/content/dam/geometrixx-outdoors/activities/jcr:content/folderThumbnail"
+        def parentNode = createNode(thumbnailFile, false, JcrConstants.NT_FILE,
+                [createNode("${thumbnailFile}/jcr:content", true, JcrConstants.NT_UNSTRUCTURED)])
+
+        when:
+        NodeProtos.Node nodeProto = new JcrNodesProcessor().process(parentNode)
+
+        then:
+        nodeProto.name == thumbnailFile
+        nodeProto.properties.propertyList.first().value.stringValue == JcrConstants.NT_FILE
+        nodeProto.mandatoryChildNodeList.size() == 1
+        nodeProto.mandatoryChildNodeList.first().name == "${thumbnailFile}/jcr:content"
+        nodeProto.mandatoryChildNodeList.first().properties.propertyList.first().name == JcrConstants.JCR_PRIMARYTYPE
+        nodeProto.mandatoryChildNodeList.first().properties.propertyList.first().value.stringValue == JcrConstants.NT_UNSTRUCTURED
+    }
+
+    def "Skip mandatory child node nt:resource as parent has already process"() {
+        given:
+        def imageFile = "/content/geometrixx-outdoors/jcr:content/image/file/jcr:content/dam:thumbnails/dam:thumbnail_600.png"
+        def childrenNode = createNode("${imageFile}/jcr:content", true, JcrConstants.NT_RESOURCE)
+
+        when:
+        NodeProtos.Node nodeProto = new JcrNodesProcessor().process(childrenNode)
+
+        then:
+        nodeProto == null
+    }
+
+    def "Example not found and extreme case where grand child node is also required"() {
+        given:
+        def thumbnailFile = "/content/dam/geometrixx-outdoors/activities/jcr:content/folderThumbnail"
+        def parentNode = createNode(thumbnailFile, false, JcrConstants.NT_FILE,
+                [createNode("${thumbnailFile}/jcr:content", true, JcrConstants.NT_RESOURCE,
+                    [createNode("${thumbnailFile}/jcr:content/metadata", true, JcrConstants.NT_UNSTRUCTURED)])])
+
+        when:
+        NodeProtos.Node nodeProto = new JcrNodesProcessor().process(parentNode)
+
+        then:
+        nodeProto.name == thumbnailFile
+        nodeProto.properties.propertyList.first().value.stringValue == JcrConstants.NT_FILE
+        nodeProto.mandatoryChildNodeList.size() == 1
+        nodeProto.mandatoryChildNodeList.first().name == "${thumbnailFile}/jcr:content"
+        nodeProto.mandatoryChildNodeList.first().properties.propertyList.first().name == JcrConstants.JCR_PRIMARYTYPE
+        nodeProto.mandatoryChildNodeList.first().properties.propertyList.first().value.stringValue == JcrConstants.NT_RESOURCE
+
+        nodeProto.mandatoryChildNodeList.first().mandatoryChildNodeList.first().name == "${thumbnailFile}/jcr:content/metadata"
+        nodeProto.mandatoryChildNodeList.first().mandatoryChildNodeList.first().properties.propertyList.first().name == JcrConstants.JCR_PRIMARYTYPE
+        nodeProto.mandatoryChildNodeList.first().mandatoryChildNodeList.first().properties.propertyList.first().value.stringValue == JcrConstants.NT_UNSTRUCTURED
+    }
+
+    private JcrNode createNode(String path, boolean isMandatory, String primaryType, Collection<JcrNode> children = []) {
+        def nodeDefinition = isMandatory ? mandatoryNodeDefinition() : nonMandatoryNodeDefinition()
+
+        def childDefinitions = children.collect {it.getDefinition()} as List<NodeDefinition>
+
+        def node = Mock(JcrNode) {
+            getPath() >> path
+            getDefinition() >> nodeDefinition
+            getProperties() >> propertyIterator(primaryTypeProperty(primaryType))
+            getPrimaryNodeType() >> Mock(NodeType) {
+                getChildNodeDefinitions() >> childDefinitions.toArray()
+            }
+        }
+
+        children.each { JcrNode child ->
+            child.getParent() >> node
+        }
+        node.getNodes() >> new NodeIteratorAdapter(children.iterator())
+
+        return node
+    }
+
+    private NodeDefinition mandatoryNodeDefinition() {
+        return Mock(NodeDefinition) {
+            isMandatory() >> true
+        }
+    }
+
+    private NodeDefinition nonMandatoryNodeDefinition() {
+        return Mock(NodeDefinition) {
+            isMandatory() >> false
+        }
+    }
+
+    private JcrProperty primaryTypeProperty(String propertyValue) {
+        return Mock(JcrProperty) {
+            getType() >> STRING
+            getName() >> JcrConstants.JCR_PRIMARYTYPE
+            getValue() >> new ValueImpl(propertyValue)
+        }
+    }
+
+    private static PropertyIterator propertyIterator(JcrProperty... properties) {
+        new PropertyIteratorAdapter(properties.iterator())
+    }
 }


sikarwarvimal:
cc: @pgoodrich

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +89,53 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
             }
         }
 
-        final List<JcrProperty> properties = jcrNode.properties.toList()
-        Node.Builder nodeBuilder = Node.newBuilder()
+        // Hierarchy node logic
+        if(isHierarchyNode(jcrNode) || true) {
```
pgoodrich: @sikarwarvimal Remove `|| true` :)

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +89,53 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
+            // It checks if the parent has required primary type as "nt:hierarchyNode"
+            // then skip it as the node has already been processed by parent node
+            if(isRequiredHierarchyChildNode(jcrNode)) {
```
pgoodrich: Can simplify this to:
```groovy
if (isRequiredHierarchyChildNode(jcrNode)) {
  return null;
}
else if (isHierarchyNode(jcrNode)) {
  return buildNode(jcrNode, getRequiredChildNodes(jcrNode))
}
```

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: Shoot, i was trying something for testing so the flow will drop in here.

sikarwarvimal: I will remove it

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +89,53 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
+    }
 
+    private static Collection<JcrNode> getRequiredChildNodes(JcrNode jcrNode) {
+        // Here we chose not to handle the edge case of a required child hierarchy node requiring a grandchild node
+        // because no node structure exists like that within AEM.
+        return jcrNode.getNodes()
+                .findAll{JcrNode childJcrNode -> isRequiredHierarchyChildNode(childJcrNode)}
+    }
+
+    private static boolean isRequiredHierarchyChildNode(JcrNode node) {
```
pgoodrich: I thought about this and this should probably be `isRequiredChildNode(...)` so we don't imply it needs to be a hierarchy node.

grabbit/src/test/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriterSpec.groovy
```diff
@@ -36,25 +37,45 @@ class JcrNodesWriterSpec extends Specification {
...
+            .setName(JcrConstants.JCR_LASTMODIFIED)
             .setType(STRING)
             .setValue(NodeProtos.Value.newBuilder().setStringValue("Date"))
             .build()
         propertiesBuilder.addProperty(aProperty)
         nodeBuilder.setProperties(propertiesBuilder.build())
+
+        // Child Node jcr:content
+        NodeProtos.Node.Builder childNodeBuilder = NodeProtos.Node.newBuilder()
+        childNodeBuilder.setName("/$JcrConstants.JCR_CONTENT")
```
pgoodrich: Should be full path to jcr:content node

grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
```diff
@@ -29,7 +27,7 @@ import static com.twcable.jackalope.JcrConstants.NT_FILE
 import static javax.jcr.PropertyType.LONG
 import static javax.jcr.PropertyType.STRING
 
-@Subject(JcrNodesProcessor)
+//@Subject(JcrNodesProcessor)
```
pgoodrich: Can you explain why this was removed?

sikarwarvimal: No need to remove, i guess i was trying something with test case that we were facing.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: I think the method name is more appropriate as we are checking both `required` and `hierarchy`. I am not sure if there would be any condition or not but we cannot use the same method incase of non `hierarchy` `required` child node.

sikarwarvimal:
Ready for testing and review after the change.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +90,51 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
             }
         }
 
-        final List<JcrProperty> properties = jcrNode.properties.toList()
-        Node.Builder nodeBuilder = Node.newBuilder()
+        // It checks if the node parent has required primary type as "nt:hierarchyNode"
+        // and then skip it as the node has already been processed by parent node
+        if(isRequiredHierarchyChildNode(jcrNode)) {
```
jbornemann: Wondering if we should encapsulate this logic within JCRNodeDecorator, instead of having JcrNodesProcessor inspect node instances.  Similar to ProtoNodeDecorator, JCRNodeDecorator should "know" how to write itself to proto format, as well as perform any necessary inspection on it's underlying inner node.    

It would spring board enthusiasm to get toProperty()'s ugly switch statement dissolved in another effort, in a similar manner of empowering our objects.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +90,51 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
+            return buildNode(jcrNode, null)
+        }
+    }
+
+    private Node buildNode(JcrNode jcrNode, Collection<JcrNode> childNodes) {
+        final Node.Builder nodeBuilder = Node.newBuilder()
         nodeBuilder.setName(jcrNode.path)
+        nodeBuilder.setProperties(buildProperties(jcrNode))
+        childNodes?.each {
+            nodeBuilder.addChildNode(buildNode(it, getRequiredChildNodes(it)))
```
jbornemann: Did we decide for a recursive flavor of algorithm?  I originally prompted this due to possibility of an nt:hierarchy's only child node being another "required" CH+ nt:hierarchy node, but last I left the conversation it seemed like we were going to leave this remote possibility alone.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +90,51 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
+        // because no node structure exists like that within AEM.
+        return jcrNode.getNodes()
+                .findAll{JcrNode childJcrNode -> isRequiredHierarchyChildNode(childJcrNode)}
+    }
 
+    private static boolean isRequiredHierarchyChildNode(JcrNode node) {
+        return node.getDefinition().isMandatory() && node.getParent().getDefinition().requiredPrimaryTypes.any{ it.toString() == JcrConstants.NT_HIERARCHYNODE }
+    }
+
+    private static boolean isHierarchyNode(JcrNode node) {
```
jbornemann: I really appreciate your style here of breaking the algorithm into functional, easily consumable pieces.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +90,51 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
+    }
+
+    private static Collection<JcrNode> getRequiredChildNodes(JcrNode jcrNode) {
+        // Here we chose not to handle the edge case of a required child hierarchy node requiring a grandchild node
+        // because no node structure exists like that within AEM.
+        return jcrNode.getNodes()
+                .findAll{JcrNode childJcrNode -> isRequiredHierarchyChildNode(childJcrNode)}
+    }
 
+    private static boolean isRequiredHierarchyChildNode(JcrNode node) {
```
jbornemann: Is there another situation where a node is mandatory?  Would

```groovy
private static boolean isRequiredHierarchyChildNode(JcrNode node) {
    node.getDefinition().isMandatory()
}
```
suffice?  If there is, should we be applying this to those "other situations"

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +90,51 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
+        return node.getDefinition().isMandatory() && node.getParent().getDefinition().requiredPrimaryTypes.any{ it.toString() == JcrConstants.NT_HIERARCHYNODE }
+    }
+
+    private static boolean isHierarchyNode(JcrNode node) {
+        try {
+            return node.isNodeType(NodeType.NT_HIERARCHY_NODE)
+        } catch (RepositoryException var3) {
```
jbornemann: Gah.  @JCR 2.0 API : "RepositoryException - if an error occurs."

jbornemann: Is this a paranoid try/catch; or should we assume this is a common JCR failure, and that the Jackrabbit team was feeling uncreative in their exception semantics this day? :)

grabbit/src/main/proto/node.proto
```diff
@@ -9,6 +9,7 @@ option java_outer_classname = "NodeProtos";
 message Node {
     required string name = 1;
     required Properties properties = 2;
+    repeated Node childNode = 3;
```
jbornemann: What do people think of calling this "mandatoryChildNode"?  Just to hint that it's not *all* child nodes being sent in a packet

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -24,12 +24,15 @@ import com.twcable.grabbit.proto.NodeProtos.Properties
 import com.twcable.grabbit.proto.NodeProtos.Property
 import groovy.transform.CompileStatic
```
jbornemann: Spec that covers this functionality sever-side?

jbornemann:
This is looking really good so far!  Exciting!

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
pgoodrich: @jbornemann isNodeType(..) throws a Repository Exception so we have to handle it here :)

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: There has been an issue with server side logic so currently it has been removed and added a internal jira ticket. I have pasted the link on internal grabbit channel.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: The recursive algorithm goes up to the node till it will keep finding `nt:hierarchy parent + required child node` so even there will be more child nodes but if it does not satisfy the same then those will be not processed.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: `isRequiredHierarchyChildNode` method checks `nt:hierarchy parent + required child node` condition to identify the correct node and that is what we should be looking. This is more defensive check as it avoids any other non-hierarchical mandatory condition, also i am not aware of any such condition but not taking any chance.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: I am fine moving the `Building Nodes` logic to `JCRNodeDecorator ` if that is the intention of `Decorator` class.

sikarwarvimal: But if you you follow the design pattern then logic is more suited for `Builder pattern or Factory pattern` rather than `Decorator Pattern`. @jdigger ?

sikarwarvimal: Even i thinking `Command Pattern` could be fit in as `HieraricalNodeCommand` and `NonHieraricalNodeCommand` to `execute`

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann: Just spoke with Jim around some findings.  It doesn't look like it's really nt:hierarchyNode types we are looking for.  If you go to your CQ's /crx/de/exportnodetype.jsp ; and navigate to nt:hierarchyNode, you'll see that it is essentially an empty semantic abstract type.  You can also find other unstructured nodes that have required mandatory children, such as rep:system (as an example).  

Rather than looking for nt:hierarchyNodes, we may need to see if a node has mandatory children/names via http://www.day.com/maven/jsr170/javadocs/jcr-2.0/javax/jcr/nodetype/NodeDefinition.html#getRequiredPrimaryTypes() and http://www.day.com/maven/jsr170/javadocs/jcr-2.0/javax/jcr/nodetype/NodeDefinition.html#getRequiredPrimaryTypeNames()

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann: Those patterns are not mutually exclusive

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann: I understand what it does.  What I'm asking is if this is useful if all known node types require *zero* or more (the zero ore more part being critical, as opposed to 1 or more) other mandatory nt:hierarchyNode types.  Part of Jim's algorithm is just to satisfy these nodes when we get to them.

This would be necessary if a node type were to exist such that it required 1 or more children of type nt:hierarchyNode

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann: I believe defined interactions in a setup/given are effectively ignored.  Try moving to when: block

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: You can give the same argument for `Decorator` pattern. Though in this case `Builder pattern and Factory pattern` are creational patterns which can be used together for complex object or separately for simple one. Regarding `Command` pattern which encapsulate a command request as an object can also be explored. But creational patterns are more appropriate.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: As per the ticket and my understanding is that we are dealing only nt:hierarchy node logic with mandatory child. If you remember the given example was type `nt:file` which is a hierarchy node and required child `jcr:content`. Currently master code has only exception for `nt:file` that has been replaced:
https://github.com/TWCable/grabbit/pull/49/files#diff-0783fe4e036efd5e6de0cc48eb601a35L102

If we want to incorporate other node details then certainly we can do on a separate PR.

grabbit/src/main/proto/node.proto
pgoodrich: :+1:

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann: Indeed, but you said it yourself.  Those patterns are creational patterns - JcrNodeFactory wouldn't accurately describe what it *is*, or *does*.  If all the class did was create JCR nodes, then absolutely; but what it *is*, and *does*; is it wraps (or decorates) a JCR node, and gives JCR nodes extra behavior.  The argument I don't think can be made for the command pattern yet, but maybe it will evolve to that

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann: Looking at nt:hierarchyNode's does not address this issue we are trying to solve.  It's a simple change to bring the solution to full fruition

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: current implemented logic is for 0 or more. Let me know if you find otherwise.

grabbit/src/main/proto/node.proto
sikarwarvimal: I am fine with the suggested change.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: You are mixing two things together `what it is, and does`. What `it is` refers to `Structural design patterns` and what `it does` is a `Behavioral design patterns`. Decorator pattern is a `Structural design pattern` which `Add responsibilities to objects dynamically`.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: I guess to incorporate the logic to handle `ALL` mandatory child nodes we need to `get rid of` all the checks about the `hierarchy node` and apply simple logic that if node has mandatory child then process that no matter what type of the node it is.
I can certainly do that but `need quick help` with testing as there will be logic change from the `earlier basic premise about the nt:hierarchy logic`

jbornemann: Will
```groovy
node.getDefinition().getRequiredPrimaryNodeTypes().find { it.mandatory } 
```
do it?  The documentation is a bit vague

jbornemann: "Gets the minimum set of primary node types that the child node must have...."

jbornemann: http://www.day.com/maven/jsr170/javadocs/jcr-2.0/javax/jcr/nodetype/NodeDefinition.html#getRequiredPrimaryTypes()

pgoodrich: Should just be able to use `node.getDefinition().isMandatory()`

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann: You're right, but alas I still think JcrNodeDecorator describes what the class does, and is,  well.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: Yes you just need `node.getDefinition().isMandatory()` as currently we are doing inside the code.

pgoodrich: I think @jbornemann's point is that we need to check the parent node to see if it requires child mandatory nodes or else we will end up checking parsing every group of children of every node for mandatory.

sikarwarvimal: I guess this simple logic will take care of all conditions.
```
        // Skip all required node as it already been processed by parent node
        if(isRequiredNode(jcrNode)) {
            return null;
        } else {
            // Build parent node 
            return buildNode(jcrNode, getRequiredChildNodes(jcrNode))
        }
```
```
    private Node buildNode(JcrNode jcrNode, Collection<JcrNode> childNodes) {
        final Node.Builder nodeBuilder = Node.newBuilder()
        nodeBuilder.setName(jcrNode.path)
        nodeBuilder.setProperties(buildProperties(jcrNode))
        childNodes?.each {
            nodeBuilder.addChildNode(buildNode(it, getRequiredChildNodes(it)))
        }
        return nodeBuilder.build()
    }
```

jbornemann:
I did some playing around in the AEM groovy console.  Looks like : "node.getPrimaryNodeType().getChildNodeDefinitions().findAll { it.mandatory }.size() > 0" should effectively be the replacement logic for "Is a nt:hiearchyNode"

sikarwarvimal:
@jbornemann You do not need this logic anymore. I have updated the logic and tested the PR. Now the all mandatory child nodes are handled and processed by the parent. Now the logic is much simpler and straight forward.

This PR is ready for testing and review.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
```diff
@@ -93,28 +84,17 @@ class JcrNodesWriter implements ItemWriter<ProtoNode>, ItemWriteListener {
...
-            * This will always be overridden by the actual jcr:data value as that will be the next thing received
-            */
-            jcrContentNode.setProperty(JCR_DATA, new BinaryValue("temp".bytes))
-            jcrContentNode.setLastModified()
-        }
         jcrNode.setLastModified()
+        // This will processed all mandatory child nodes only
+        if(nodeProto.mandatoryChildNodeList && nodeProto.mandatoryChildNodeList.size() > 0) {
+            for(ProtoNode childNode: nodeProto.mandatoryChildNodeList) {
+                writeToJcr(childNode, session)
```
jbornemann: setLastModified() on these?

sikarwarvimal: check line number 89

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +86,57 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
+        return nodeBuilder.build()
+    }
 
+    /**
+     * Identify all required child nodes
+     * @param JCR Node
+     * @return list of required child nodes
+     */
+    private static Collection<JcrNode> getRequiredChildNodes(JcrNode jcrNode) {
+        return jcrNode.getNodes()
```
jbornemann: one line?

sikarwarvimal:
Spock Test case is failing and i am looking into it.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +86,57 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
         }
 
-        final List<JcrProperty> properties = jcrNode.properties.toList()
-        Node.Builder nodeBuilder = Node.newBuilder()
+        // Skip the node as it already been processed by parent node
+        if(isRequiredNode(jcrNode)) {
+            return null;
+        } else {
+            // Build parent node
+            return buildNode(jcrNode, getRequiredChildNodes(jcrNode))
```
jbornemann:
```groovy
final requiredChildNodes = hasMandatoryChildNodes(jcrNode) ? getRequiredChildNodes(jcrNode) : null
return buildNode(jcrNode, requiredChildNodes)
...
private boolean hasMandatoryChildNodes(jcrNode) {
    jcrNode.getPrimaryNodeType().getChildNodeDefinitions().find { it.mandatory }
}
```
so we don't have to iterate over all of a node's child nodes if we don't have to (as is the case most of the time)

sikarwarvimal: I think it is a good call. Modified a bit:

```
    private static Collection<JcrNode> getRequiredChildNodes(JcrNode jcrNode) {
        return hasMandatoryChildNode(jcrNode) ? jcrNode.getNodes().findAll{JcrNode childJcrNode -> isRequiredNode(childJcrNode)} : null
    }

    private static boolean hasMandatoryChildNodes(JcrNode jcrNode) {
        return jcrNode.getPrimaryNodeType().getChildNodeDefinitions().any { ((ItemDefinition)it).isMandatory() }
    }
```

sikarwarvimal:
Ready for final round of testing and review.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jdigger: Interactions are commonly defined in a `setup:` block. There are some edge-case rules, but generally you declare them wherever it makes it easiest to read.

https://spockframework.github.io/spock/docs/1.0/interaction_based_testing.html#_where_to_declare_interactions

jdigger:
Please add more test cases to handle the various cases described in the algorithm in #34.

sagarsane:
Couple test cases we can try out :

`/content/geometrixx-outdoors/jcr:content/image/file/jcr:content/dam:thumbnails/dam:thumbnail_600.png/jcr:content`

Another one : 

`/content/dam/geometrixx-outdoors/activities/jcr:content/folderThumbnail/jcr:content` where "folderThumbnail"'s "jcr:content" is of type "nt:unstructured"

(These paths should be present in all instances as its Geometrixx)

sikarwarvimal:
Thanks! @sagarsane I tested both cases and able to sync both.
Note i deleted `/content/dam/geometrixx-outdoors` and `/content/geometrixx-outdoors` nodes on publish and synced from author.

jdigger:
To the issue of not being able to use Spock's stubbing/mocking for the Jackalope-generated classes, the solution is simply to use a little Proxy magic with Groovy's Closures:

```groovy
private static <T> T wrapInstance(Object src, Map<String, Closure> closureMap) {
    return (T) Proxy.newProxyInstance(src.class.classLoader, src.class.interfaces, new AbstractInvocationHandler() {
        protected Object handleInvocation(Object proxy, Method method, Object[] args) throws Throwable {
            Closure closureMethod = closureMap.get(method.name)
            return (closureMethod != null) ? closureMethod.call(args) : method.invoke(src, args)
        }
    })
}
```

Then you can use a line like
```groovy
JcrNode hierarchyNode = wrapInstance(fakeNodeBuilder.build(), [isNodeType: { true }])
```
to swap out behavior.

jdigger:
Please at least construct a test case (or cases) as described in GH-34 to validate things.
Better would be to include the situations Sagar pointed out as well.

sikarwarvimal:
Yes, currently i am working on the test cases.

grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
```diff
@@ -69,4 +75,136 @@ class JcrNodesProcessorSpec extends Specification {
         propertyString.type == STRING
     }
 
+    def "Can marshall a JCR Node with mandatory child node to a Protobuf Message"() {
```
sikarwarvimal: I have added test case for a node with mandatory child. It was fun to write the test case as there were few interesting challenges while writing. Could not use Jackalope APIs also i had to write some custom iterators for testing.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
```diff
@@ -93,28 +84,17 @@ class JcrNodesWriter implements ItemWriter<ProtoNode>, ItemWriteListener {
...
-            * If jcr:data is not part of that chunk then you will get a constraint violation exception
-            * To get around that, just add an empty binary jcr:data here with a "temp" value
-            * This will always be overridden by the actual jcr:data value as that will be the next thing received
-            */
-            jcrContentNode.setProperty(JCR_DATA, new BinaryValue("temp".bytes))
-            jcrContentNode.setLastModified()
-        }
         jcrNode.setLastModified()
+        // This will processed all mandatory child nodes only
+        if(nodeProto.mandatoryChildNodeList && nodeProto.mandatoryChildNodeList.size() > 0) {
```
sagarsane: Think `if(nodeProto.mandatoryChildNodeList)` will be sufficient? (for null and non empty check?)

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
```diff
@@ -95,7 +95,7 @@ class ProtoNodeDecorator {
         property.values.valueList.each { ProtoValue value ->
             if (node.canAddMixin(value.stringValue)) {
                 node.addMixin(value.stringValue)
-                log.info "Added mixin ${value.stringValue} for : ${node.name}."
+                log.debug "Added mixin ${value.stringValue} for : ${node.name}."
```
sagarsane: too much noise? ;-)

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +87,65 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
+        }
+    }
+
+    /**
+     * Build node and "only" mandatory child nodes
+     * @param JCR Node
+     * @param Mandatory childNodes
+     * @return Node Proto node
+     */
+    private Node buildNode(JcrNode jcrNode, Collection<JcrNode> childNodes) {
```
sagarsane: This is minor .. but the method for marshalling JCR type to Protobuf type was called `toProperty()` below ... could keep the same convention and name this `toNode()` and similarly .. `toProperties()` below instead of `buildProperties()` ...

grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
```diff
@@ -69,4 +75,136 @@ class JcrNodesProcessorSpec extends Specification {
...
+        then:
+        nodeProto != null
+        nodeProto.name == "default.groovy"
+        // Verified parent node as nt:file which has mandatory child node
+        nodeProto.properties.propertyList.any { it.name == JcrConstants.JCR_PRIMARYTYPE && it.value.stringValue == JcrConstants.NT_FILE}
+        // Verified child node "default.groovy/jcr:content", which is mandatory child node and a property type as nt:resource
+        nodeProto.mandatoryChildNodeList.any { it.name == "default.groovy/jcr:content" && it.properties.propertyList.any { it.name == JcrConstants.JCR_PRIMARYTYPE && it.value.stringValue == JcrConstants.NT_RESOURCE}}
+    }
+
+    // Need to implement the NodeIterator as neither the Mock nor the Jackalope was helpful
```
sagarsane: I don't think you need to implement your own Node & Property iterators ... you can leverage the Jackrabbit Commons' `NodeIteratorAdapter` and `PropertyIteratorAdapter` here (that can convert normal iterators to `NodeIterator` or `PropertyIterator`)

Example : in our case : 

```groovy
hierarchyNode.getNodes() >> new NodeIteratorAdapter([childMandatoryNode].iterator())
//...
childMandatoryNode.properties >> new PropertyIteratorAdapter([childPrimaryProperty].iterator())
//...
//etc
```

sagarsane: Also, I am wondering why you weren't able to use the Jackalope APIs (did you find any implementation missing in Jackalope APIs?)

sagarsane:
I ran some tests on 5.6.1 instance (using Geometrixx content) and it looks pretty good so far :+1:

grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
sikarwarvimal: Cool, i couldn't find the PropertyIterator or NodeIterator implementation but now i have replaced with Jackrabbit's Adapter implementation.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: I think `build` as part of method name is more apt as it based on the action of the method. Probably `toProperty` can be renamed as `setupProperty` or `readProperty` to make it more appropriate.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
sikarwarvimal: Yes it was tooooo much to handle during testing :)

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
sikarwarvimal: I was sure about the not null check but not the empty one so kept it explicit.

sikarwarvimal:
I have updated the test case. Now everyone has reported +1 for regression testing so waiting for +1 on CR and once I will get the CR then the branch will be ready for merge.

sikarwarvimal:
@jdigger this task https://travis-ci.org/TWCable/grabbit/jobs/72839045 is failing on `travis-ci` though is not failing locally. Any thoughts?

jbornemann:
Any thoughts on moving some of the new JcrNodesProcessor, and similar logic to JCRNodeDecorator?  Instead of inspecting, and building proto messages within JcrNodesProcessor, we could empower JCRNodeDecorator to write itself to proto format, and provide decorated functionality for inspecting node/properties. 

Should be a straightforward move.

jbornemann:
A checksum might be wrong for a cache entry, I cleared the cache to see if that helps..

jbornemann:
Looks like that did it

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann: Ah okay.  Yeah, playing a bit in a sandbox, that does seem to be the case.  Maybe I've just had bad luck with edge cases ;)

sikarwarvimal:
Okay, as you have already created the helper class `JCRNodeDecorator` i will move those methods there.

sikarwarvimal:
@jbornemann as I started looking to move out the code from `JcrNodesProcessor` to `JCRNodeDecorator`  I realized that It can be done with a separate PR because I do not want to leave few inherent methods inside `JcrNodesProcessor` like `toProperty` method and moved few out. I think it would be a better approach when we move out as one cohesive unit together. Also in terms of testing this PR feels good so want go with it as after the change again regression testing is required.

grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
```diff
@@ -69,4 +77,62 @@ class JcrNodesProcessorSpec extends Specification {
...
+        nodeProto.properties.propertyList.any {
+            it.name == JcrConstants.JCR_PRIMARYTYPE && it.value.stringValue == JcrConstants.NT_FILE
+        }
+        // Verified child node "default.groovy/jcr:content", which is mandatory child node and a property type as nt:resource
+        nodeProto.mandatoryChildNodeList.any {
+            it.name == "default.groovy/jcr:content" && it.properties.propertyList.any {
+                it.name == JcrConstants.JCR_PRIMARYTYPE && it.value.stringValue == JcrConstants.NT_RESOURCE
+            }
+        }
+    }
```
jdigger: Rather than that, which is awfully hard to follow the setup, a cleaner and more Spock-like way would be
```groovy
def "Can marshall a JCR Node with mandatory child node to a Protobuf Message"() {
    given:
    def mandatoryChildNodeDefinition = Mock(NodeDefinition) {
        isMandatory() >> true
    }

    // Setup parent node as nt:file which has mandatory child node
    def hierarchyNode = Mock(JcrNode) {
        getPath() >> "default.groovy"
        getDefinition() >> Mock(NodeDefinition)
        getProperties() >> new PropertyIteratorAdapter([Mock(JcrProperty) {
            getType() >> STRING
            getName() >> JcrConstants.JCR_PRIMARYTYPE
            getValue() >> new ValueImpl(JcrConstants.NT_FILE)
        }].iterator())
        getPrimaryNodeType() >> Mock(NodeType) {
            getChildNodeDefinitions() >> [mandatoryChildNodeDefinition]
        }
    }

    // Setup child node "default.groovy/jcr:content", which is mandatory child node and a property type as nt:resource
    def childMandatoryNode = Mock(JcrNode) {
        getPath() >> "default.groovy/jcr:content"
        getDefinition() >> mandatoryChildNodeDefinition
        getProperties() >> new PropertyIteratorAdapter([Mock(JcrProperty) {
            getType() >> STRING
            getName() >> JcrConstants.JCR_PRIMARYTYPE
            getValue() >> new ValueImpl(JcrConstants.NT_RESOURCE)
        }].iterator())
        getParent() >> hierarchyNode
        getPrimaryNodeType() >> Mock(NodeType) {
            getChildNodeDefinitions() >> [Mock(NodeDefinition)]
        }
    }

    hierarchyNode.getNodes() >> new NodeIteratorAdapter([childMandatoryNode].iterator())

    when:
    NodeProtos.Node nodeProto = new JcrNodesProcessor().process(hierarchyNode)

    then:
    nodeProto.name == "default.groovy"
    nodeProto.properties.propertyList.first().value.stringValue == JcrConstants.NT_FILE
    nodeProto.mandatoryChildNodeList.size() == 1
    nodeProto.mandatoryChildNodeList.first().name == "default.groovy/jcr:content"
}
```
The `then:` section is also greatly simplified to only test what we care about.

Or the test could be simplified even more simply with
```groovy
def "Can marshall a JCR Node with mandatory child node to a Protobuf Message "() {
    given:
    def hierarchyNode = createNode("default.groovy", false, JcrConstants.NT_FILE,
            [createNode("default.groovy/jcr:content", true, JcrConstants.NT_RESOURCE)])

    when:
    NodeProtos.Node nodeProto = new JcrNodesProcessor().process(hierarchyNode)

    then:
    nodeProto.name == "default.groovy"
    nodeProto.properties.propertyList.first().value.stringValue == JcrConstants.NT_FILE
    nodeProto.mandatoryChildNodeList.size() == 1
    nodeProto.mandatoryChildNodeList.first().name == "default.groovy/jcr:content"
}

NodeDefinition mandatoryNodeDefinition() {
    return Mock(NodeDefinition) {
        isMandatory() >> true
    }
}


NodeDefinition nonMandatoryNodeDefinition() {
    return Mock(NodeDefinition) {
        isMandatory() >> false
    }
}


JcrProperty primaryTypeProperty(String propertyValue) {
    return Mock(JcrProperty) {
        getType() >> STRING
        getName() >> JcrConstants.JCR_PRIMARYTYPE
        getValue() >> new ValueImpl(propertyValue)
    }
}


static PropertyIterator propertyIterator(JcrProperty... properties) {
    new PropertyIteratorAdapter(properties.iterator())
}


JcrNode createNode(String path, boolean isMandatory, String primaryType, Collection<JcrNode> children = []) {
    def nodeDefinition = isMandatory ? mandatoryNodeDefinition() : nonMandatoryNodeDefinition()

    def childDefinitions = children.collect {it.getDefinition()} as List<NodeDefinition>

    def node = Mock(JcrNode) {
        getPath() >> path
        getDefinition() >> nodeDefinition
        getProperties() >> propertyIterator(primaryTypeProperty(primaryType))
        getPrimaryNodeType() >> Mock(NodeType) {
            getChildNodeDefinitions() >> childDefinitions.toArray()
        }
    }

    children.each { JcrNode child ->
        child.getParent() >> node
    }
    node.getNodes() >> new NodeIteratorAdapter(children.iterator())

    return node
}
```
The advantage here is that the feature itself becomes trivial to not only create the node structure, but also to understand it.

sikarwarvimal: Thanks @jdigger i will update as suggested.

jdigger:
Now that setting up a testing node structure for unit tests is relatively easy, please make sure to construct a test case (or cases) as described in GH-34 and the situations Sagar pointed out to validate the "known" paths.

sikarwarvimal:
@jdigger i have added more test cases for different mandatory nodes. Also all the test cases from GH-34 are not valid anymore as the implementation as been further simplified. Now there is no more logic difference between `hierarchy nodes` and `non-hierarchy nodes`. The new logic is completely based on the `mandatory child node`. So now all the `mandatory child nodes` are processed together with the parent without identifying the type of the parent.

grabbit/src/main/proto/node.proto
```diff
@@ -9,6 +9,7 @@ option java_outer_classname = "NodeProtos";
 message Node {
     required string name = 1;
     required Properties properties = 2;
+    repeated Node mandatoryChildNode = 3;
```
jdigger: `mandatoryChildNode` should be plural since there can be more than one.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -88,10 +87,65 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
             }
         }
 
-        final List<JcrProperty> properties = jcrNode.properties.toList()
-        Node.Builder nodeBuilder = Node.newBuilder()
+        // Skip the node as it already been processed by parent node
```
jdigger: Minor grammar: "Skip this node because it has already been processed by its parent"

grabbit/src/main/proto/node.proto
pgoodrich: I think `mandatoryChildNode` is correct here due to repeated qualifier (see `repeated Property property` below).  Otherwise our generated methods will be get, set, addMandatoryChildNodes(...) plural when in reality they only handle one node at a time.

sikarwarvimal: also so make it multivalue the protobuffer generates method name and variable with `List` in it as `mandatoryChildNodeList` based on `repeated` flag.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
sikarwarvimal: Will update this.

jdigger:
+1 CR :+1:

PR #55, opened 2015-08-05
Work to solve edge case around server writing mandatory nodes

jbornemann:
To support changes @sikarwarvimal made, this will fix an edge case if a configuration "/. path" targets a node with mandatory children that are not jcr:content.  

Some shifting around of things as well...

@sagarsane @jdigger @sikarwarvimal @masroormohammed


grabbit/src/main/groovy/com/twcable/grabbit/server/services/NodeWithMandatoryIterator.groovy
```diff
@@ -0,0 +1,90 @@
...
+
+import javax.jcr.Node as JcrNode
+
+/**
+ * Custom Node Iterator that will iterate over a rootNode node and its immediate children
+ * taking into account cases where a node (rootNode node or any of the children) has any number of
+ * mandatory subnodes
+ */
+@CompileStatic
+final class NodeWithMandatoryIterator implements Iterator<JcrNode> {
```
sikarwarvimal: `RootNodeWithMandatoryIterator`?

grabbit/src/main/groovy/com/twcable/grabbit/server/services/NodeWithMandatoryIterator.groovy
```diff
@@ -0,0 +1,90 @@
...
+    private JCRNodeDecorator rootNode
+    private Iterator<JCRNodeDecorator> immediateChildren
+    private Iterator<JCRNodeDecorator> mandatoryWriteNodes
+
+
+    public NodeWithMandatoryIterator(JcrNode root) {
+        this.rootNode = new JCRNodeDecorator(root)
+        this.doneRoot = false
+        //Get all immediate children that are not mandatory write nodes.  We will handle those by iterating over mandatoryWriteNodes
+        immediateChildren = rootNode.getImmediateChildNodes().findAll { !it.isRequiredNode() }.iterator() as Iterator<JCRNodeDecorator>
```
sikarwarvimal: helper method `JCRNodeDecorator.getNonMandatoryChildrens(root)`?

grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
```diff
@@ -35,12 +38,27 @@ class JCRNodeDecorator {
     JCRNode innerNode
 
 
+    //Evaluated in a lazy fashion
+    private Collection<JCRNodeDecorator> immediateChildNodes
```
sagarsane: Do we need this private variable? It looks like we are directly calling `getImmediateChildNodes()` when needed. So the method can just return the `immediateChildNodes` ?

grabbit/src/main/groovy/com/twcable/grabbit/server/services/NodeWithMandatoryIterator.groovy
```diff
@@ -0,0 +1,90 @@
...
+        //Calls a tail recursive method, gathering all required nodes
+        mandatoryWriteNodes = addToMandatoryIterator(rootNode, []).iterator()
```
sagarsane: I am not sure if this was discussed offline; but was wondering will this handle the complete sub-tree under `/content/x/jcr:content` for example where `jcr:content` is required child for `x` but its subtree can be something like 

`/content/x/jcr:content/a/b/c` -- a is required, but `b` and `c` arent .. similarly ..
`/content/x/jcr:content/d/e/f/g` -- where none inside `jcr:content` are `required` technically but `semantically` they make sense for say a `cq:Page` ?

sagarsane: The prior `JcrContentRecursiveIterator` handled this case.

grabbit/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
jbornemann: For performance, I'd like immediate children to be evaluated in a lazy fashion (as below).

grabbit/src/main/groovy/com/twcable/grabbit/server/services/NodeWithMandatoryIterator.groovy
jbornemann: Sure

grabbit/src/main/groovy/com/twcable/grabbit/server/services/NodeWithMandatoryIterator.groovy
jbornemann: Sure

grabbit/src/main/groovy/com/twcable/grabbit/server/services/NodeWithMandatoryIterator.groovy
jbornemann: Not the complete subtree.  Just the root node, it's immediate children, and any nodes required to save the subtree.

Thinking we should leave it up to a node's definition to define what a node's semantic meaning is.  In your example, jcr:content is apart of a cq:Page's semantic meaning, but AEM does not define the types of (a,b,c,etc) as defining parts.  

That way, we do exactly what a /. path implies, without making any assumptions.

sagarsane: I don't think it makes sense to get just the root 'jcr:content' and not its complete subtree (if present). Without the subtree, the cq:Page doesn't really mean anything. The subtree is what defines the page. I think it should continue to send the subtree if present. I understand that this can be considered a 'special case' but I think this makes more sense.

And not just about cq:Page thats just one example, there can be other cases(maybe in DAM)

What do others think?

jbornemann: Well, in the case of a configuration path "foo/somepage/." , you're saying you just want to grab the page + page properties.  I don't think it makes sense to create compexlity, and surprise results for a few node types;  Just let the node definition define what is important.

If one wanted everything from a single page, why not explicitly express what you want;  i.e do the regular recursive syntax, and exclude subpages you don't want?

jbornemann: Not to mention, if you DID just want a page + page properties;  making it behave in a special way would break user expectations

PR #56, opened 2015-08-06
Use request.remoteUser on both Client and Server to create JCR Session

sagarsane:
...

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Moreover, by doing a continuous stream, we avoid the latency issues. Depending o
 
 Below details AEM version support for the various releases of Grabbit.  
 ```
-v4.0.0 - AEM 6.1/CQ 5.6
+v4.0.x - AEM 6.1
 v3.0.0 - CQ 5.6
 v2.0.1 - CQ 5.6
 v2.0.0 - CQ 5.6
--- a/grabbit/src/integrationTest/groovy/com/twcable/grabbit/server/services/ServerServiceSpec.groovy
+++ b/grabbit/src/integrationTest/groovy/com/twcable/grabbit/server/services/ServerServiceSpec.groovy
@@ -72,7 +72,7 @@ class ServerServiceSpec extends Specification {
 
         when:
         //This will also actually execute the Batch Job internally
-        syncServerService.getContentForRootPath("/default.groovy", (Collection<String>)Collections.EMPTY_LIST, "", mockServletOutputStream)
+        syncServerService.getContentForRootPath("admin", "/default.groovy", (Collection<String>)Collections.EMPTY_LIST, "", mockServletOutputStream)
 
         then:
         mockServletOutputStream != null
@@ -85,7 +85,7 @@ class ServerServiceSpec extends Specification {
 
         when:
         //This will also actually execute the Batch Job internally
-        syncServerService.getContentForRootPath("/default.groovy", ["/default.groovy/jcr:content"] as Collection<String>, "", mockServletOutputStream)
+        syncServerService.getContentForRootPath("admin", "/default.groovy", ["/default.groovy/jcr:content"] as Collection<String>, "", mockServletOutputStream)
 
         then:
         mockServletOutputStream != null
--- a/grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
+++ b/grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
@@ -42,8 +42,9 @@ class ClientBatchJob {
     public static final String WORKFLOW_CONFIGS = "workflowConfigIds"
     public static final String HOST = "host"
     public static final String PORT = "port"
-    public static final String USERNAME = "username"
-    public static final String PASSWORD = "password"
+    public static final String SERVER_USERNAME = "serverUsername"
+    public static final String SERVER_PASSWORD = "serverPassword"
+    public static final String CLIENT_USERNAME = "clientUsername"
     public static final String CONTENT_AFTER_DATE = "contentAfterDate"
     public static final String DELETE_BEFORE_WRITE = "deleteBeforeWrite"
 
@@ -99,8 +100,9 @@ class ClientBatchJob {
     @CompileStatic
     static class CredentialsBuilder {
         final ServerBuilder parentBuilder
-        String username
-        String password
+        String clientUsername
+        String serverUsername
+        String serverPassword
 
 
         CredentialsBuilder(ServerBuilder parentBuilder) {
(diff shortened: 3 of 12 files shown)


sagarsane:
@jdigger

sagarsane:
@jdigger @jbornemann @sikarwarvimal please take a look. I will create a branch off v3.0.0 as well and patch that version as well.

Should I update the version to `v4.0.1` and the other to `v3.0.1` ?

jbornemann:
Mind providing some reasoning behind the change?

4.0.0  is backwards compatible, so changing 3.0.0 isn't necessary

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
```diff
@@ -42,8 +42,9 @@ class ClientBatchJob {
     public static final String WORKFLOW_CONFIGS = "workflowConfigIds"
```
jbornemann: Not really related to your change, but wondering in retrospect if a builder really makes sense if you only ever have one client, and are building in one fashion.

Seems every change to this file endures lots of development overhead to maintain pattern.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
```diff
@@ -42,8 +42,9 @@ class ClientBatchJob {
...
     public static final String HOST = "host"
     public static final String PORT = "port"
-    public static final String USERNAME = "username"
-    public static final String PASSWORD = "password"
+    public static final String SERVER_USERNAME = "serverUsername"
+    public static final String SERVER_PASSWORD = "serverPassword"
+    public static final String CLIENT_USERNAME = "clientUsername"
```
jbornemann: Can we assume one UN/PW combination?  Seems like it would keep things simpler.

jazeren1:
@jbornemann are you positive 4.0 is backwards compatible and the change isn't needed in the 3.0 stream?  It was my understanding that 3.0 was for CQ 5.6.1 and 4.0 was for AEM 6.1.  If it's truly backwards compatible, then way do we have a 3.0 and a 4.0 released?

grabbit/src/integrationTest/groovy/com/twcable/grabbit/server/services/ServerServiceSpec.groovy
```diff
@@ -72,7 +72,7 @@ class ServerServiceSpec extends Specification {
 
         when:
         //This will also actually execute the Batch Job internally
-        syncServerService.getContentForRootPath("/default.groovy", (Collection<String>)Collections.EMPTY_LIST, "", mockServletOutputStream)
+        syncServerService.getContentForRootPath("", "/default.groovy", (Collection<String>)Collections.EMPTY_LIST, "", mockServletOutputStream)
```
jazeren1: @sagarsane I must admin this looks a little odd to me.  I see you added a new param for this function so the first one should be username, but I'm not sure what is going on here.  I assume this test is passing, but I'm not entirely sure why :wink:

jdigger:
Grabbit v4 is not backwards compatible with AEM 5.6 **_unless you already have Guava installed_**. In addition to some different bindings to various libraries, where full regression testing hasn't been done to make sure that something at runtime wouldn't cause an issue, Guava is the most obvious issue. AEM 6 has Guava v15 pre-installed, so it's removed from the list of bundles included in the Grabbit v4 package to make sure there's no potential conflicts.

grabbit/src/integrationTest/groovy/com/twcable/grabbit/server/services/ServerServiceSpec.groovy
sagarsane: Yeah I should pass in `"admin"` instead of `""` :) For an `empty` username I believe Jackalope internally defaults to anonymous user that's why the test passes.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
sagarsane: Because a `Grabbit Server` user only needs to `read` and not write.. and `Client` user needs to write, I think that would not be a sound assumption.

jdigger: As an example, consider the case where the client is your local development instance and the server is production. You local instance is probably admin/admin, but the production server had better not be using those credentials... So the client needs to know the name of the user to write to its JCR as, and the credentials to use when connecting to the server for retrieving the data.

sagarsane:
Updated branch with CR changes and documentation. Ready for more review and test.

README.md
```diff
@@ -82,6 +82,7 @@ directory in order for Grabbit to be built from source.
 6. Enter the path to the Grabbit config you created in step. Screenshot of how it will look like : ![alt text](assets/GrabbitConfig.jpg)
 7. Once the Grabbit content sync is kicked off, you will get a confirmation of the kicked off jobs like : ![alt text](assets/GrabbitComplete.jpg)
 
+**Note:** The client and server side credentials entered above will be used as the credentials when creating the JCR session on both Client and Server.
```
sagarsane: Can this be worded any better?

jdigger: To be honest, I don't think the "Note:" is needed.

sagarsane: Ok :-) I thought it would be better if I was explicit about this.

jdigger: Normally I'm all for explicit. In this case I think the context makes it clear enough and this just muddies the water.

sagarsane: Okay cool.

sagarsane:
Removed the "Note" from the README.

grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy
sagarsane: Maybe. Though I still like the way we build out the Client (& Server) batch job. I think because there are so many things to configure, it may still make sense.

jdigger:
The code looks good, but can someone please give this some "practical" tests?  (e.g., Have protected and "open" content, try connecting anonymously, as admin, as a user that doesn't have access to the protected content, etc.)

sagarsane:
I did the following test : 

I created a user and gave it read access only to   
`/var/eventing` ... and 
`/var/eventing/jobs`
and had that user ask for `/var/eventing` ..  
  
As expected, I only saw `/var/eventing/jobs` tree sent by the server. On the client side, I also created a user who did NOT have write access to `/var/eventing/jobs` ... here too; as expected, the user was not able to write to this location above.

I had previously already tested for `admin` user and that works as expected as well.

Also renamed the Documentation commit with a better commit message.

I will back-port the main fix to `3.x.x` branch as well.

sagarsane:
Updated `3.x.x` branch with this fix https://github.com/TWCable/grabbit/commits/3.x.x

Issue #39, opened 2015-07-01
Delta content syncs


jdigger:
The feature is "implemented" but it never got properly tested.

https://github.com/TWCable/grabbit/search?utf8=%E2%9C%93&q=deltaContent

It's based on last successful jobexecution "endTime" for a path on the client.

https://github.com/TWCable/grabbit/blob/844b0d13abec0a693990f323cf3e6a2693e366a1/grabbit/src/main/groovy/com/twcable/grabbit/servlets/GrabbitServlet.groovy#L65

https://github.com/TWCable/grabbit/blob/844b0d13abec0a693990f323cf3e6a2693e366a1/grabbit/src/main/groovy/com/twcable/grabbit/client/batch/ClientBatchJob.groovy#L200-L215 

Needs cleanup, testing and documentation.

egenz:
To set deltaContent, you may need to add the line:
`"deltaContent" : true,`
if it is not already present in your grabbit-publish-dev.json file.

The deltaContent setting determines what is new content based on the 'afterDateString' (seen in the second link above) and synchs only those nodes to the client.

Test Cases (done by synching with Grabbit from AEM 6.1 author instance to publish instance):
* Adding or copying a component to a page:
 * Works with deltaContent as expected, updates with Grabbit.

* Modifying a component with no created date (only "LastModified" date) [e.g. an Image in a Slide Component]: 
 * Changing LastModified Date trips deltaContent, updates w/ Grabbit
 * (as expected) just changing the image on the node, without updating LastModified, doesn't work.

* Modifying a component with a 'created' and LastModified date [e.g. Button component]
 * Changing LastModified date *doesn't* work, running grabbit does not update the page
 * Changing 'created' date does work, and the instance updates with grabbit. (even if lastModified stays the same)
 * May be due to logic within grabbit which returns 'date' as either 'created' OR 'lastModified', so these are not both included in the information grabbit knows about a particular node. When only the 'created' date is understood as the date, the 'afterDateString' won't realize the node has been updated.

* Deleting a component:
 * Doesn't work, as expected.

* Testing a larger path (170,000 nodes):
 * Still correctly only updates a single node.

* Multiple Changes (ie, changing created date on a button and modified date on rich text):
 * Only successfully syncs components which have properly updated 'lastModified' or 'created' dates as experienced above

PR #58, opened 2015-08-13
Delta content

egenz:
testing and documenting deltaContent setting as detailed in https://github.com/TWCable/grabbit/issues/39

With this setting on, grabbit should now appropriately sync only content which was added or changed since the previous grabbit sync date. Tested with various node properties and sizes / scopes of changes and paths.

to test, run a grabbit sync to a local instance with deltaContent as 'true'. Add or change components on a path to be synced, then sync again and make sure only those nodes are synced across (the grabbit-start-monitor will show how many nodes were copied). 

@jdigger @sikarwarvimal @masroormohammed for CR/Testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ A `json` configuration file of following format is used to configure Grabbit.
     "serverPassword" : "<password>",
     "serverHost" : "some.other.server",
     "serverPort" : "4502",
+    "deltaContent" : true,
     "pathConfigurations" :  [
         {
             "path" : "/content/someContent",
@@ -133,6 +134,8 @@ A `json` configuration file of following format is used to configure Grabbit.
     * __path__: The path to recursively grab content from.
 
 #### Optional fields
+
+* __deltaContent__: boolean, where ```false``` has grabbit sync & overwrite all content nodes regardles of date properties, and ```true``` syncs only 'delta' or changed content. Changed content is determined by comparing either of the jcr:lastModified, cq:lastModified, or jcr:created Date with the Date when the path was successfully synced as recorded by the Grabbit Client. Thus, deltaContent will only work with paths that remained the same between two runs. Nodes without any date properties will also be synced, so as not to miss new content of unstructured nodes.
 
 Under "path configurations"
 
--- a/grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
+++ b/grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
@@ -77,11 +77,7 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
             final Date afterDate = DateUtil.getDateFromISOString(contentAfterDate)
             log.debug "ContentAfterDate received : ${afterDate}. Will ignore content created or modified before the afterDate"
             final date = getDate(jcrNode)
-            if (!date) {
-                //we want delta content but node doesn't have any date property to compare with. So we ignore it
-                return null
-            }
-            if (date.before(afterDate)) {
+            if (date && date.before(afterDate)) { //if there are no date properties, we treat nodes as new
                 log.debug "Not sending any data older than ${afterDate}"
                 return null
             }
@@ -165,14 +161,14 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
      */
     private static Date getDate(JcrNode jcrNode) {
         final String CQ_LAST_MODIFIED = "cq:lastModified"
-        if (jcrNode.hasProperty(JCR_CREATED)) {
-            return jcrNode.getProperty(JCR_CREATED).date.time
-        }
-        else if (jcrNode.hasProperty(JCR_LASTMODIFIED)) {
+        if (jcrNode.hasProperty(JCR_LASTMODIFIED)) {
             return jcrNode.getProperty(JCR_LASTMODIFIED).date.time
         }
         else if (jcrNode.hasProperty(CQ_LAST_MODIFIED)) {
             return jcrNode.getProperty(CQ_LAST_MODIFIED).date.time
+        }
+        else if (jcrNode.hasProperty(JCR_CREATED)) {
+            return jcrNode.getProperty(JCR_CREATED).date.time
         }
         return null
     }
--- a/grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
+++ b/grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
@@ -34,6 +34,10 @@ import static com.twcable.jackalope.JCRBuilder.node
 import static com.twcable.jackalope.JCRBuilder.property
 import static javax.jcr.PropertyType.LONG
 import static javax.jcr.PropertyType.STRING
+import org.joda.time.DateTime
+import javax.jcr.NodeIterator
+import spock.lang.Shared
+import spock.lang.Unroll
 
 @Subject(JcrNodesProcessor)
 class JcrNodesProcessorSpec extends Specification {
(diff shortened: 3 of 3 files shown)


grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -178,6 +187,23 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
+    private static Date getModifiedDate(JcrNode jcrNode) {
+        final String CQ_LAST_MODIFIED = "cq:lastModified"
+        if (jcrNode.hasProperty(JCR_LASTMODIFIED)) {
+            return jcrNode.getProperty(JCR_LASTMODIFIED).date.time
+        }
+        else if (jcrNode.hasProperty(CQ_LAST_MODIFIED)) {
+            return jcrNode.getProperty(CQ_LAST_MODIFIED).date.time
+        }
+        return null
+    }
```
sagarsane: Hey @egenz , there isn't much difference between this method and https://github.com/egenz/grabbit/blob/deltaContent/grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy#L169-L187 this method. Could you please explain why this is needed?

pgoodrich: @egenz I think we can just reverse the order of getDate so it checks CQ_LAST_MODIFIED and JCR_LASTMODIFIED before JCR_CREATED.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -77,13 +77,22 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
             final Date afterDate = DateUtil.getDateFromISOString(contentAfterDate)
             log.debug "ContentAfterDate received : ${afterDate}. Will ignore content created or modified before the afterDate"
             final date = getDate(jcrNode)
+            final modifiedDate = getModifiedDate(jcrNode)
             if (!date) {
                 //we want delta content but node doesn't have any date property to compare with. So we ignore it
                 return null
             }
             if (date.before(afterDate)) {
```
pgoodrich: At some point would like to replicate nodes without a JCR_CREATED, JCR_LASTMODIFIED, or CQ_LAST_MODIFIED if they are mandatory nodes and their parent as been modified.

@jdigger @sikarwarvimal

jdigger:
A few of the tests can be replaced taking a data-driven approach like
```groovy
@Shared
DateTime oldDate = new DateTime(2015, 8, 4, 15, 24, 34, 961)

@Unroll
def "Process a node with old date properties: #propList"() {
    when:
    JcrNodesProcessor jcrNodesProcessor = new JcrNodesProcessor()
    jcrNodesProcessor.setContentAfterDate(oldDate.plusDays(1).toString())
    NodeProtos.Node nodeProto = jcrNodesProcessor.process(aJcrNode)
    aJcrNode.properties.toList()

    then:
    nodeProto == null //not copying old data

    where:
    aJcrNode << [
            node("testComponent",
                    property(JcrConstants.JCR_PRIMARYTYPE, JcrConstants.NT_UNSTRUCTURED),
                    // no date is treated as "infinitely old"
            ).build(),
            node("testComponent",
                    property(JcrConstants.JCR_PRIMARYTYPE, JcrConstants.NT_UNSTRUCTURED),
                    property(JcrConstants.JCR_CREATED, oldDate.toCalendar(Locale.default)),
            ).build(),
            node("testComponent",
                    property(JcrConstants.JCR_PRIMARYTYPE, JcrConstants.NT_UNSTRUCTURED),
                    property(JcrConstants.JCR_LASTMODIFIED, oldDate.toCalendar(Locale.default))
            ).build(),
            node("testComponent",
                    property(JcrConstants.JCR_PRIMARYTYPE, JcrConstants.NT_UNSTRUCTURED),
                    property(JcrNodesProcessor.CQ_LAST_MODIFIED, oldDate.toCalendar(Locale.default))
            ).build(),
            node("testComponent",
                    property(JcrConstants.JCR_PRIMARYTYPE, JcrConstants.NT_UNSTRUCTURED),
                    property(JcrConstants.JCR_CREATED, oldDate.toCalendar(Locale.default)),
                    property(JcrConstants.JCR_LASTMODIFIED, oldDate.toCalendar(Locale.default))
            ).build(),
            node("testComponent",
                    property(JcrConstants.JCR_PRIMARYTYPE, JcrConstants.NT_UNSTRUCTURED),
                    property(JcrConstants.JCR_CREATED, oldDate.toCalendar(Locale.default)),
                    property(JcrNodesProcessor.CQ_LAST_MODIFIED, oldDate.toCalendar(Locale.default))
            ).build(),
    ]
    propList = aJcrNode.properties.toList().findAll {it.name != JcrConstants.JCR_PRIMARYTYPE}.collectEntries { [(it.name): new DateTime(it.value.date.time.time)]}
}
```

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jdigger: That's a great test to add: Does the node aggregator kick in under such a circumstance?

egenz: Working on this now. The way the code is written, it seems like this should be addressed by lines 92-97, where the parent and children are processed together, but I'm testing now to see if that's the case.

grabbit/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
egenz: Interestingly, I tried swapping them the first time I was looking for a solution to this, and it caused grabbit to fail... upon trying again though I must've done something different, as my tests are passing. I'll remove the getModifiedDate method unless I run into issues again.

egenz:
@sagarsane @jdigger @pgoodrich @jbornemann @sikarwarvimal 

Ready again for CR & Testing

grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
```diff
@@ -191,4 +197,151 @@ class JcrNodesProcessorSpec extends Specification {
...
+    @Unroll
+    def "Process a node with updated content and date properties: #propList"() {
+        when:
+        JcrNodesProcessor jcrNodesProcessor = new JcrNodesProcessor()
+        jcrNodesProcessor.setContentAfterDate(oldDate.plusDays(1).toString())
+        NodeProtos.Node nodeProto = jcrNodesProcessor.process(aJcrNode)
+        aJcrNode.properties.toList()
+
+        then:
+        nodeProto != null //not copying old data
```
jdigger: The comment is incorrect here: c&p error

jbornemann:
@egenz We are going to want the README updated under the configuration section with how to use this functionality

jbornemann:
I'll give it a test whirl!

README.md
```diff
@@ -134,6 +135,8 @@ A `json` configuration file of following format is used to configure Grabbit.
 
 #### Optional fields
 
+* __deltaContent__: boolean, where ```false``` has grabbit sync & overwrite all content nodes regardles of date properties, and ```true``` syncs only 'delta' or changed content and leaves unchanged content alone (with deltaContent on, syncs will be significantly faster).
```
sagarsane: May be we should also add the fact that the `changed content` is calculated by comparing either of the `jcr:lastModified`, `cq:lastModified` or `jcr:created` Date with the Date when the `path` was successfully synced as recorded by the Grabbit Client.

In other words, we should be explicit that the deltaContent will only make sense if the paths stay the same for subsequent runs.

That is : path - `/a/b/c/d` finished successfully and then we ran `/a/b/c/d` again with `deltaContent : true` then that will work but if we run `/a/b/c/d/e` the second time, `deltaContent : true` will not work since there was no job run for `/a/b/c/d/e` before.

egenz:
@sagarsane @jdigger @pgoodrich @jbornemann @sikarwarvimal 

updated readme and how deltaContent works with dateless nodes, please test & CR again.

jbornemann:
@egenz should we capture the "no date on node" scenario in a spec feature?

egenz:
@jbornemann line 317 of the spec should be testing that

grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
```diff
@@ -191,4 +195,151 @@ class JcrNodesProcessorSpec extends Specification {
...
+                        property(JcrConstants.JCR_PRIMARYTYPE, JcrConstants.NT_UNSTRUCTURED),
+                        property(JcrConstants.JCR_CREATED, oldDate.toCalendar(Locale.default)),
+                        property("cq:lastModified", oldDate.toCalendar(Locale.default))
+                ).build(),
+        ]
+        propList = aJcrNode.properties.toList().findAll {it.name != JcrConstants.JCR_PRIMARYTYPE}.collectEntries { [(it.name): new DateTime(it.value.date.time.time)]}
+    }
+
+    @Shared
+    DateTime currentDate = new DateTime()
```
sagarsane: Does this need to be shared? This seems to be used only in the test below, the `currentDate` in the 1st test looks like a local variable.

sagarsane:
Apart from the last comment, looks good to me CR wise +1. Have not tested, but looks like this has already been tested a bit....

grabbit/src/test/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessorSpec.groovy
egenz: @sagarsane it seems to need to be shared... tried a couple different ways of declaring it and tests fail.

jbornemann:
Testing proved successful for me!  Test +1

Could you squash your commits @egenz , and I'll merge this in?

jdigger:
Merged

PR #62, opened 2015-08-19
Fixed issues that crept in with deleteBeforeWrite

jbornemann:
DeleteBeforeWrite is effectively broken on 3.0.x/4.0.x.  Some how breaking Spring configuration changes got in after test (Maybe IDEA was trying to be *too* smart during refactor)

Also removed some tasks from using SimpleAsyncTaskExecutioner to address a concern of @sagarsane 's in #48 .  

Tested to some extent, and it seems to look good now.  

@jdigger @masroormohammed @sikarwarvimal


grabbit/src/main/resources/META-INF/spring/client-deleteBeforeWrite-steps.xml
```diff
@@ -23,7 +23,7 @@
         <bean id="deleteBeforeWriteDecider" class="com.twcable.grabbit.client.batch.steps.workspace.DeleteBeforeWriteDecider" />
         <bean id="deleteBeforeWriteTasklet" class="com.twcable.grabbit.client.batch.steps.workspace.DeleteBeforeWriteTasklet" scope="step">
             <constructor-arg name="jobPath" value="#{jobParameters[path]}" />
-            <constructor-arg name="excludePaths" value="#{jobParameters[path]}" />
-            <constructor-arg name="slingRepository" value="slingRepository" />
+            <constructor-arg name="excludePaths" value="#{jobParameters[excludePaths]}" />
+            <constructor-arg name="slingRepository" ref="slingRepository" />
```
sagarsane: This can be done in a different branch but Now that `deleteBeforeWrite` is not in a different thread, we can use the session managed in `ThreadLocal` for consistency; thus removing the need to have this `slingRepository` reference.

sagarsane:
Apart from the comment (which is not critical), +1 CR. Not tested. I am guessing we will back port this to `3.x.x` branch as well, and release new versions for both?

jbornemann:
Yeah, that's what I was thinking too

PR #67, opened 2015-09-02
Update README to add information about predefined loggers

sagarsane:
Small README change @jdigger @jbornemann please take a look.


README.md
```diff
@@ -216,6 +216,10 @@ __Sample of a real Grabbit Job status__
 
 ![alt text](assets/ExampleGrabbitStatus.jpg)
 
+Two loggers are predefined for Grabbit. One for Grabbit Server and the other for Grabbit Client.
```
jbornemann: Maybe too much info?  I think the majority of people reading the README are users, rather than Grabbit developers.

sagarsane: Its still worth adding I think; but may be under "advanced monitoring" or some heading maybe?

jdigger: Users would care a great deal about the log files...

The `README.md` file is being rendered from the context of https://github.com/TWCable/grabbit, so you can use a relative path to `blob/master/grabbit/src/main/content/SLING-INF/content/apps/grabbit/config/org.apache.sling.commons.log.LogManager.factory.config-com.twcable.grabbit.client.batch.xml`

sagarsane:
Updated README and also added details about #71.

Issue #69, opened 2015-09-02
Be able to specify "continue on error"


jdigger:
Grabbit is currently -- by design -- quite fragile. If there's any bad data, incomplete transmission of data, etc. it is blatantly obvious that something went wrong with no possibility of "silent" errors. Better to fail fast and spectacularly than to let issues hide.

That said, sometimes "99%" is good enough, especially when dealing with development systems and/or unreliable networks. There should be a way to say for a particular job that it should log any errors it gets, but otherwise just keep going.

mzgubin:
@jdigger this is kind of what I've worked on with my PR https://github.com/TWCable/grabbit/pull/214
That said, I think this is something we should be able to configure. Going to create an enhancement request for this.

PR #73, opened 2015-09-10
WIP: ContinueOnError

jdigger:
For GH-69


gradle.properties
```diff
@@ -22,7 +22,7 @@ guava_version = 15.0
 httpcomponents_version = 4.3.4
 httpcomponents_client_version = 4.3.6
 httpcomponents_core_version = 4.3.2
-jackalope_version = 2.0.0
+jackalope_version = 3.0.1
```
sagarsane: We can also update the README#Library Attribution section to update the version we use.

Issue #63, opened 2015-08-20
Try to remove need for Adobe JARs


jdigger:
Since Adobe's public repo does not have functional 6.1 jars, can we either remove the need to use the workflow jar altogether or potentially compile against an older version so that CI works again?

sagarsane:
The workflow jar is needed for the `Automated Workflow management` feature we have in grabbit. I suppose we could compile against older version.

Issue #38, opened 2015-07-01
WorkflowManagerSpec "Should manage multithreading correctly" fails intermittently 


jbornemann:
This happens locally, and on CI.  Need to figure out if the test is flawed, or if legitimately have multithreading issues.

jbornemann:
#97

PR #82, opened 2015-09-29
#63 Use earlier version of workflow library so CI jobs can perform

jbornemann:
@sagarsane @jdigger 

- Removed unused Gradle files
- Compile against Adobe Nexus available cq-workflow v5.6.6

My testing run through passed, but may want to have someone else try


jbornemann:
#63

sagarsane:
I built the project several times and verified that the cq-workflow-console jar is correctly excluded from the build and that the project gets correctly compiled. I think this looks good +1

gradle.properties
```diff
@@ -13,7 +13,7 @@ commons_collections_version = 3.2.1
 commons_io_version = 1.4
 commons_lang_version = 3.3.2
 commons_text_version = 1.1.8
-cq_workflow_console_version = 5.8.2
+cq_workflow_console_version = 5.6.6
```
jdigger: Should probably add a comment acknowledging that it's an "old" version, but needed to compile because of the change in obfuscation policies.

jdigger:
Yaaaay!  Passing CI builds!

Ironically the latest build has an error, but that's related to #38 

+1 with just the minor comment

jbornemann:
Sure thing!

Issue #86, opened 2015-10-12
deltaContent functionality at overall config level instead of path level


egenz:
deltaContent is set by overall configuration in the json file instead of on a path-by-path basis. We'd like to specify which paths are using this setting instead of the entire sync.

jdigger:
To both not break existing config and give a nice feature, I propose that the global delta flag stay, but it gets "applied" to each of the paths unless a path overrides it.

That way if you want delta for 90% of your paths but not one or two (or vice versa), you can do so easily.

PR #88, opened 2015-10-15
Delta content configs

egenz:
Changing deltaContent configs as per this issue

https://github.com/TWCable/grabbit/issues/86

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,8 @@ A `JSON` configuration file of following format is used to configure Grabbit.
                     "/etc/workflow/launcher/config/update_asset_create",
                     "/etc/workflow/launcher/config/dam_xmp_nested_writeback",
                     "/etc/workflow/launcher/config/dam_xmp_writeback"
-                ]
+                ],
+            "deltaContent" : false
         }
     ]
 }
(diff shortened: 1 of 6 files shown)


egenz:
@jdigger @sikarwarvimal @masroormohammed @cculb For testing & review.

sagarsane:
It looks like this branch needs to be synced with master.

cffried:
@egenz  we were seeing that deltas were writing but not actually doing anything. you can test this by changing a single node on a page and running the sync. You will see the node written but when refreshing nothing happens

egenz:
@sagarsane synced branch
@cffried I tested this by syncing grabbit locally using several configurations with the deltaContent flag, and successfully had syncs happening, including changing a single node on one page, then running the sync. The change was synced over to my other instance.

jazeren1:
@egenz was your test using this branch or without this branch?  What @cffried was seeing was outside of this branch, but it could be that this branch helps with that issue.

egenz:
The specific tests I was mentioning were on this branch, but I also tested deltaContent before it went in, and had success. Could the problem have been with configuration? there seems to have been some confusion over where deltaContent is set. This branch should also fix that.

masroormohammed:
So ran a couple of tests on this branch and everything looks to be copied over correctly. 

+1 Testing

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -73,7 +73,14 @@ class GrabbitConfiguration {
             def excludePaths = nonEmptyCollection(config["excludePaths"] as Collection<String>, errorBuilder)
             def workflowConfigIds = config["workflowConfigIds"] as Collection<String> ?: [] as Collection<String>
             def deleteBeforeWrite = boolVal(config, 'deleteBeforeWrite')
-            new PathConfiguration(path, excludePaths, workflowConfigIds, deleteBeforeWrite)
+            def boolean pathDeltaContent
```
jdigger: Trivial: no need for `def` when declaring the datatype.

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -73,7 +73,14 @@ class GrabbitConfiguration {
...
+            if (!config.containsKey('deltaContent')) {            //if: no pathDeltaContent value
+                pathDeltaContent = configMap.get('deltaContent')  //get global deltaContent value
+            }
+            else {
+                pathDeltaContent = config.get('deltaContent')
+            }
```
jdigger: FWIW, this is Groovy, so you have the Elvis-operator:

```groovy
// use the global deltaContent if a local one doesn't exist
boolean pathDeltaContent = config.get('deltaContent') ?: configMap.get('deltaContent')
```

egenz: this is way snazzier. updating.

egenz:
@sagarsane @cffried @jazeren1 @jdigger updated as per CR, please review.

jdigger:
+1 CR

README.md
```diff
@@ -161,13 +162,14 @@ pathConfigurations :
...
+* __deltaContent__: boolean, ```true``` syncs only 'delta' or changed content. Changed content is determined by comparing one of a number of date properties including jcr:lastModified, cq:lastModified, or jcr:created Date with the last successful Grabbit sync date. Nodes without any of previously mentioned date properties will always be synced even with deltaContent on.  Most common throughput bottlenecks are usually handled by delta sync for cases such as large DAM trees; but if your case warrants a more fine tuned use of delta sync, you may consider adding mix:lastModified to nodes not usually considered for exclusion, such as extremely large unstructured trees.
```
sagarsane: I am wondering what was changed here :-) since there doesn't seem to be any change in the wording..

sagarsane:
Apart from one minor comment, +1 CR.

README.md
egenz: I don't think there was a change, not sure why it's showing one. However, upon further thought I think I should probably add some things because there's been some confusion over how deltaContent works -- i'll do that now.

egenz:
I have made a small change to the readme documentation, should warrant a quick look to make sure it makes sense. Besides that, I believe this should be ready for merging.

masroormohammed:
+1 readme aspect

PR #89, opened 2015-10-16
WIP - Grabbit Server should not ignore JCR_LASTMODIFIED from sending to client

sagarsane:
@jdigger 

cc @jazeren1 FYI this is likely the fix for the "outdated" issue.

There are unit tests failing right now but wanted to open up for visibility.


jbornemann:
Is there an issue # associated with this?  

What's the background?

jdigger:
Is this still "In progress"?

sagarsane:
No its not. I'll close it. This was a need at the time but I think we found a work-around. I'll close this and keep the branch around in case its needed at any point.

Issue #112, opened 2016-06-15
Allow passing the batch size as part of job configuration


jdigger:
It's currently [hard coded at 1,000 nodes in DefaultClientService](https://github.com/TWCable/grabbit/blob/fdba912/grabbit/src/main/groovy/com/twcable/grabbit/client/services/impl/DefaultClientService.groovy#L42)

The appropriate batch size can vary dramatically, whether it's thousands of tiny nodes or a few massive nodes. Should allow passing the batch size in as a configuration of the job/path.

jdigger:
It looks like it may also be set in `grabbit/src/main/resources/META-INF.spring/client-batch-job.xml`. This should be consolidated (or at least better documented) as part of this.

jbornemann:
I think we can close this one right @jdigger ? Reopen if I am wrong

PR #115, opened 2016-06-28
112 added option to specify commit interval in config file

viveksachdeva:
Issue Id:
https://github.com/TWCable/grabbit/issues/112

Changes:
Added Option to specify commit interval in config file. If not specified, defaults to 100.

@sagarsane @jdigger @jbornemann @jazeren1

### DIFF
--- a/grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
+++ b/grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
@@ -42,6 +42,8 @@ class GrabbitConfiguration {
 
     final long transactionID
 
+    final static int DEFAULT_BATCH_SIZE = 100
+
 
     private GrabbitConfiguration(@Nonnull String user, @Nonnull String pass, @Nonnull String host,
                                  @Nonnull String port, boolean deltaContent,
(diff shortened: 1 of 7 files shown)


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/115" target="_blank">ReviewNinja

jdigger:
It doesn't look like it's specified at the job/path level so that nodes under `/etc` can be treated differently than those under `/content/dam`. Does the way the Spring Batch configuration work not allow doing so?

sagarsane:
@viveksachdeva could you also see if you could add / update `GrabbitConfigurationSpec` to account for this new feature?

viveksachdeva:
@jdigger : It can be done at job level and if not present there, can be defaulted to transaction level..

viveksachdeva:
@sagarsane : Sure.. will add those

viveksachdeva:
@sagarsane @jdigger : I have updated code to pick batch size from pathConfiguration. So the priority order is:
- path configuration
- transaction
- default(100)
Test cases have also been updated..

viveksachdeva:
@sagarsane @jdigger @masroormohammed @jazeren1 @jbornemann 

This PR is ready for review..

jbornemann:
Thanks! I think this looks good!

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -152,6 +156,19 @@ class GrabbitConfiguration {
...
+            log.debug "Input doesn't contain ${key} for a boolean value. Will default to null"
+        }
+        Integer integerValue
+        try {
+            integerValue = configMap.get(key) as Integer
+        }
+        catch (NumberFormatException numberFormatException){
+            log.debug "${key} is not a number"
+        }
+        return integerValue
```
jdigger: What does this return if `config.get(key)` contains "fooble"?

jdigger: !resolve

jdigger:
Other than the return value ambiguity, it looks good.

grabbit/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
viveksachdeva: !resolved

viveksachdeva:
@jdigger : Added default value 0 in case value is not a number..

viveksachdeva:
For testing:
With following sample config:
    "serverUsername" : "admin",
            "serverPassword" : "admin",
            "serverHost" : "localhost",
            "serverPort" : "4503",
            "deltaContent" : false,
            "batchSize"  : 50,
            "pathConfigurations" :  [
                {
                    "path" : "/content/a/b",
                    "batchSize"  : 150,
                },
                {
                    "path" : "/content/xyz",
                }
            ]
        }

there should be 2 jobs in the transaction:
- The one with path "/content/a/b" should have batch size 150
- The one with path "/content/xyz" should have batch size 50

If batchSize is not specified anywhere in the config, it should be 100.

Shashi-Bhushan:
:+1: Testing 

Tested with following configurations
**{
	"serverUsername" : "admin",
	"serverPassword" : "admin",
	"serverHost" : "localhost",
	"serverPort" : "7503",
	"deltaContent" : false,
	"pathConfigurations" : [
		{
			"path" : "/content/usergenerated",
			"batchSize" : 150,
		},
		{
			"path" : "/content/campaigns"
		}
	]
}**

And

**{
	"serverUsername" : "admin",
	"serverPassword" : "admin",
	"serverHost" : "localhost",
	"serverPort" : "7503",
	"deltaContent" : false,
	"batchSize" : 50,
	"pathConfigurations" : [
		{
			"path" : "/content/usergenerated",
			"batchSize" : 150,
		},
		{
			"path" : "/content/campaigns"
		}
	]
}**

jdigger:
What does a batch size of 0 do? (Or, for that matter, any non-positive number.)

viveksachdeva:
@jdigger : Negative value of batch size was not handled before.. I have pushed a commit for the same.. Negative numbers and 0 are now ignored.. The next config in the order of priority is picked.. So if job level config is negative, the one at transaction level would be used and that would fallback to default batch size of 100..

jdigger:
Perfect. +1

jdigger:
@viveksachdeva Please squash that last commit, then I'll merge it in.

jbornemann:
@viveksachdeva I think this looks good! I tested, and didn't see any issues!

I do think we will want to update the README to reflect this new configuration option

viveksachdeva:
@jdigger , @jbornemann : Done.. 👍

sagarsane:
👍

PR #133, opened 2016-07-21
Upgrade to protoc 2.6.1

jbornemann:
- This upgrades Grabbit to use the protocol buffers 2.6.1 java runtime, and provides related tooling such as protoc 2.6.1 compiler
- Message reorganization, and some new 2.6 performance features
- misc cleanup


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/133" target="_blank">ReviewNinja

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -18,41 +18,32 @@ package com.twcable.grabbit.server.batch.steps.jcrnodes
 
 import com.google.protobuf.ByteString
 import com.twcable.grabbit.DateUtil
-import com.twcable.grabbit.proto.NodeProtos
-import com.twcable.grabbit.proto.NodeProtos.Node
-import com.twcable.grabbit.proto.NodeProtos.Properties
-import com.twcable.grabbit.proto.NodeProtos.Property
+import com.twcable.grabbit.proto.NodeProtos.Node as ProtoNode
```
jbornemann: I've seen these red rectangles from time to time; but I can't seem to find any information about them. Anyone know what these are?

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -194,168 +185,25 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
-            case WEAKREFERENCE:
-                break
+        if(jcrProperty.type == BINARY) {
+            propertyBuilder.addValues(valueBuilder.setBytesValue(ByteString.copyFrom(jcrProperty.value.binary.stream.bytes)))
+        }
+        else {
+            //Other property types can potentially have multiple values
+            final Value[] values = jcrProperty.multiple ? jcrProperty.values : [jcrProperty.value] as Value[]
+            values.each { Value value ->
+                propertyBuilder.addValues(valueBuilder.setStringValue(value.string))
```
jbornemann: Looks like Jackrabbit does a good job of conforming to LSP as for as value subtypes go.

src/main/proto/node.proto
```diff
@@ -3,31 +3,22 @@ package com.twcable.grabbit.proto;
 option java_package = "com.twcable.grabbit.proto";
 option java_outer_classname = "NodeProtos";
 
-/*
```
jbornemann: None of this was necessary for conforming to the new version of protoc, but thought it would be nice to consolidate our messaging a bit, and use some of the new features. 

One thing that was sort of curious, was that I was expecting as a sort of unimportant side-effect to find a byte savings per node, and another byte savings per multi-value property with this restructure based on my understanding of the wire encoding. 

Ignoring the other stuff, if we were to send a single property with the old encoding, I would expect one byte to indicate the wire type and field value for properties (0x12), proceeded by another byte for the property header, which I thought would be removed with this PR (0x0A), then so on and so forth for property fields. 

When I did a before and after wget to /grabbit/content?path=/test/path, I received the same number of bytes before and after. Wondering if there is another optimization going on for single field embedded messages, or if I don't understand the full picture. 

Anyhoo..

.travis.yml
```diff
@@ -1,15 +1,10 @@
 language: groovy
 
-#The Travis team has approved protobuf 2.4.1 as a "safe" apt package.  2.4.1 is currently the only version available.  See full whitelist here : https://github.com/travis-ci/apt-package-whitelist
-addons:
-  apt:
-    packages:
-      - protobuf-compiler
```
jbornemann: 2.4 is the latest version available within the approved apt mirrors for travis containers, and 2.5 is the latest version of protoc you can sanely build with apt outside of containers for Ubuntu Precise Pangolan (which we probably wouldn't want to do anyway). A bunch of stdlibc++6 and related library fun; I should know, I tried and had tons of fun 🐼 . 

So..we build from source, and cache, which isn't too bad..

gradle/dependencies.gradle
```diff
@@ -91,7 +91,7 @@ dependencies {
 
     compile "org.yaml:snakeyaml:${snakeyaml_version}"
 
-    compile "org.apache.servicemix.bundles:org.apache.servicemix.bundles.protobuf-java:${servicemix_protobuf_java_version}"
+    compile "com.twc.osgi:protobuf-java:${protobuf_version}"
```
jbornemann: Couldn't find > 2.4.1 rolled within service mix, or anywhere else; so rolled our own

jbornemann:
@jdigger @sagarsane @jazeren1 ready for review

jbornemann:
@pgoodrich

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann: \<span class="pl-ii"\> 🚯

install_protoc_from_source.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env sh
+
+PROTOC_VERSION="2.6.1"
+
+if [[ ! -d "$HOME/protobuf/lib" ||  "$(protoc --version)" != *${PROTOC_VERSION}* ]]; then
```
pgoodrich: Likely edge case but what if user needs another version of protobuf installed for another application?

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
pgoodrich: Likely Github thinks the import x as y syntax is an error

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -84,81 +76,80 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
         }
     }
 
     /**
      * Build node and "only" mandatory child nodes
-     * @param JCR Node
-     * @param Mandatory childNodes
-     * @return Node Proto node
+     * @param jcrNode
+     * @param mandatoryChildNodes
```
pgoodrich: Minor: missing return in javadoc

docs/Prerequisites.adoc
```diff
@@ -3,14 +3,10 @@
 === Installing Protocol Buffers Compiler
 
 ==== For Windows
-Install the binary from https://github.com/google/protobuf/releases/download/v2.4.1/protoc-2.4.1-win32.zip and then set the `/path/to/protoc/parent` on your PATH variable.
+For Windows, the easiest way to install the compiler is to install the binary from https://github.com/google/protobuf/releases/download/v2.6.1/protoc-2.6.1-win32.zip and then set the `/path/to/protoc/parent` on your PATH variable.
```
pgoodrich: Minor: "For Windows" twice here?

pgoodrich:
On trying to run our ETC job:
```
29.07.2016 14:07:32.285 *ERROR* [clientJobLauncherTaskExecutor-4] org.springfram
ework.batch.core.step.AbstractStep Encountered an error executing step clientNam
espaceSync in job clientJob
com.google.protobuf.InvalidProtocolBufferException: Message missing required fie
lds: entry[0].uri
        at com.google.protobuf.UninitializedMessageException.asInvalidProtocolBu
fferException(UninitializedMessageException.java:81)
        at com.google.protobuf.AbstractParser.checkMessageInitialized(AbstractPa
rser.java:71)
        at com.google.protobuf.AbstractParser.parseDelimitedFrom(AbstractParser.
java:245)
        at com.google.protobuf.AbstractParser.parseDelimitedFrom(AbstractParser.
java:251)
        at com.google.protobuf.AbstractParser.parseDelimitedFrom(AbstractParser.
java:49)
        at com.twcable.grabbit.proto.NamespaceProtos$NamespaceRegistry.parseDeli
mitedFrom(NamespaceProtos.java:259)
        at com.twcable.grabbit.client.batch.steps.namespace.NamespaceSyncTasklet
.execute(NamespaceSyncTasklet.groovy:54)
        at org.springframework.batch.core.step.tasklet.TaskletStep$ChunkTransact
ionCallback.doInTransaction(TaskletStep.java:395)
        at org.springframework.transaction.support.TransactionTemplate.execute(T
ransactionTemplate.java:131)
        at org.springframework.batch.core.step.tasklet.TaskletStep$2.doInChunkCo
ntext(TaskletStep.java:267)
        at org.springframework.batch.core.scope.context.StepContextRepeatCallbac
k.doInIteration(StepContextRepeatCallback.java:77)
        at org.springframework.batch.repeat.support.RepeatTemplate.getNextResult
(RepeatTemplate.java:368)
        at org.springframework.batch.repeat.support.RepeatTemplate.executeIntern
al(RepeatTemplate.java:215)
        at org.springframework.batch.repeat.support.RepeatTemplate.iterate(Repea
tTemplate.java:144)
        at org.springframework.batch.core.step.tasklet.TaskletStep.doExecute(Tas
kletStep.java:253)
        at org.springframework.batch.core.step.AbstractStep.execute(AbstractStep
.java:195)
        at org.springframework.batch.core.job.SimpleStepHandler.handleStep(Simpl
eStepHandler.java:141)
        at org.springframework.batch.core.job.flow.JobFlowExecutor.executeStep(J
obFlowExecutor.java:64)
        at org.springframework.batch.core.job.flow.support.state.StepState.handl
e(StepState.java:60)
        at org.springframework.batch.core.job.flow.support.SimpleFlow.resume(Sim
pleFlow.java:151)
        at org.springframework.batch.core.job.flow.support.SimpleFlow.start(Simp
leFlow.java:130)
        at org.springframework.batch.core.job.flow.FlowJob.doExecute(FlowJob.jav
a:135)
        at org.springframework.batch.core.job.AbstractJob.execute(AbstractJob.ja
va:301)
        at org.springframework.batch.core.launch.support.SimpleJobLauncher$1.run
(SimpleJobLauncher.java:134)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.
java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor
.java:617)
        at java.lang.Thread.run(Thread.java:745)
```

install_protoc_from_source.sh
jbornemann: Luckily $HOME/protobuf/lib is a non-standard install point, so if someone wanted another version of protobuf installed via conventional methods, it would likely result in another binary being placed in /usr/bin or whatever. At that point, one could just provide a FQ path, or do some environment wrangling to invoke the version they want.

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann: Yeah, I deleted any JavaDoc statements that were trivial, and otherwise unhelpful. In my opinion, it's just pedantic noise. Like in this case : 

' @return Node Proto node'

However, I did put an @Nonnull annotation on the return to indicate callee/caller contract, as that stuff is important.

docs/Prerequisites.adoc
jbornemann: Hmm, one is the header title. Is the following "For Windows" redundant, should I remove it? I thought it would make things even more clear to the reader, but maybe not..

jdigger: Being immediately after the header, it feels pretty redundant.

jdigger:
I was afraid of that. Looks like it may not be binary compatible with the prior protobuf version?

install_protoc_from_source.sh
jdigger: On the point of not being a standard install path, that's not really a good thing. Better would be to put it in `/opt` or `/usr/local` or ... Including the version in the path would address the version conflicts, akin to what Homebrew does.

jbornemann: Yeah, fair enough

src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
```diff
@@ -85,17 +85,22 @@ class ProtoPropertyDecorator {
 
 
     boolean isMultiple() {
-        return hasValues()
+        return valuesCount > 1
+    }
+
+
+    ProtoValue getValue() {
+        innerProtoProperty.valuesList.first()
```
pgoodrich: @jbornemann Looks like there's some cases where there are no values present on the node which causes the overall job to fail.

29.07.2016 17:15:12.973 *ERROR* [clientJobLauncherTaskExecutor-9] org.springfram
ework.batch.core.step.AbstractStep Encountered an error executing step clientJcr
Nodes in job clientJob
java.util.NoSuchElementException: Cannot access first() element from an empty Li
st
        at org.codehaus.groovy.runtime.DefaultGroovyMethods.first(DefaultGroovyM
ethods.java:7010)
        at com.twcable.grabbit.jcr.ProtoPropertyDecorator.getValue(ProtoProperty
Decorator.groovy:93)
        at com.twcable.grabbit.jcr.ProtoPropertyDecorator.getPropertyValue(Proto
PropertyDecorator.groovy:98)

jbornemann:
Oh, @jdigger one thing I didn't think about regarding installing to /opt, or /usr/local is that these paths require su access, which generally isn't a problem for local environments, but since we are also using this script to install protoc for CI; it would mean giving up containerized (quick) builds, since sudo use is locked down to plain-ole-vm's. 

Are you okay with $HOME/lib/protoc?

src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
jbornemann: We may have to look into this deeper. A property will always have at least one value.

pgoodrich: @jbornemann Found an example in /etc/segmentation of a node with resourceType of cq/personalization/components/segmentpage that has no value for the field cq:tags.

Thinking this segment was copied from another and the tag values manually removed in the JCR.

jdigger: What can we do to be more resilient?

pgoodrich: Think we have two choices:
1. Maintain empty value
2. Truncate properties with no values

According to the JCR 2.0 spec at least, a property can have zero or more values which seems to be backed up by the node validation in CRX. https://docs.adobe.com/content/docs/en/spec/jcr/2.0/3_Repository_Model.html

jbornemann: Yeah, according to the spec, specifically multi-value properties can have an "empty" value : 

```
3.6.3 Single and Multi-Value Properties
A property may be a single-value or a multi-value property.
A single-value property, if it exists, must have a value. There is no such thing as a null value. A multi-value property can have zero or more values. Again there is no such thing as a null value, however a multi-value property can be empty, just as an array can be empty.
```

jbornemann: I didn't know this was a valid state. I'll be sure to update the code and spec to support. But yeah, I think maintain the empty value.

jbornemann: Learn new things about the spec every day 📚

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
viveksachdeva: Does that mean it automatically does toString part for all the property types?...
Moreover, Date type property was handled differently in switch case above.. Should we be handling that differently here to?

jbornemann: Yeah, it's calling getString for all types except binary values.

If you look under the hood, and using your Date example; Jackrabbit implements org.apache.jackrabbit.value.DateValue, which extends org.apache.jackrabbit.value.BaseValue, which implements the Value spec (javax.jcr.Value).  DateValue is just a nice wrapper around BaseValue, and internally does essentially the same thing we were doing before. 

BaseValue.getString() calls an internal abstract method getInternalString(), and DateValue simply implements this by calling ISO8601.format() on it's internal Date reference.

This way, you don't really need to know the subtype of Value in order to interact with getString(); you can expect consistent behavior between a DateValue, and StringValue when calling getString. Even BinaryValue.getString() creates a byte array, and interprets it as a string using the default encoding.

jbornemann:
Hey @pgoodrich when you have a chance, can you see if this clears up the issue you were running into before?

pgoodrich:
@jbornemann Works between local environments with no failures

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
viveksachdeva: so its already ISO format date.. nice 👍

jbornemann:
Great! Thanks @pgoodrich . Ready for CR

jbornemann:
@viveksachdeva do you mind taking a look at this tomorrow?

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
```diff
@@ -84,81 +76,80 @@ class JcrNodesProcessor implements ItemProcessor<JcrNode, Node> {
...
+        final transferableProperties = properties.findAll { JcrProperty jcrProperty ->
             //Before adding a property to the Current Node Proto message, check if the property
             //is Valid and if it should be actually sent to the client
-            if (isPropertyTransferable(jcrProperty)) {
-                Property property = toProperty(jcrProperty)
-                propertiesBuilder.addProperty(property)
-            }
+            isPropertyTransferable(jcrProperty)
+        }
+        return transferableProperties.collect { JcrProperty jcrProperty ->
```
viveksachdeva: rather than iterating over properties twice, can we just do findResults and add transferable check there?

viveksachdeva: Other than this, 👍 .. Will test in a while and update the PR...

viveksachdeva:
@jbornemann : Gradle build is failing for me..
Execution failed for task ':generateProto'.
> /Users/vivek/Applications/grabbit/build/extracted-protos/main: warning: directory does not exist.
  node.proto:20:5: Expected "required", "optional", or "repeated".
  node.proto:20:16: Missing field number.

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
jbornemann:
```groovy
final transferableProperties =  []
for(JcrProperty property : properties) {
  if(isPropertyTransferable(jcrProperty)) {
    transferableProperties.add(toProperty(property))
  }
}
return transferableProperties
```
Something like that? Harder to use collect, because then you have to worry about nulls in the resulting collection. Both methods are small and easy to reason about; but I sort of prefer the functional style the way it is now. Less state to keep track of. What do you think? If you prefer the above, or have another suggestion, I will change it

jbornemann:
@viveksachdeva do you have the new version of protoc compiler installed? Can check by running "protoc --version". It should be 2.6.1. Your error above is failing on new 2.6 proto syntax.

src/main/proto/node.proto
```diff
@@ -19,15 +16,20 @@ message Properties {
 message Property {
     required string name = 1;
     required int32 type = 2;
-    optional Value value = 3;
-    optional Values values = 4;
+    //essentially a C-Style Union
```
sagarsane: hmm interesting

viveksachdeva:
I ran install script that you added. I thought it would overwrite the old one...

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
viveksachdeva: I prefer the functional style too.. The problem of null can be taken care by using findResults instead of collect.. It essentially does the same thing + filter null values...

src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
```diff
@@ -85,17 +85,26 @@ class ProtoPropertyDecorator {
 
 
     boolean isMultiple() {
-        return hasValues()
```
sagarsane: @jbornemann was `hasValues()` not working for the case @pgoodrich mentioned which is why this change was needed?

viveksachdeva:
One of the tests seem to be failing:
com.twcable.grabbit.jcr.ProtoNodeDecoratorSpec > Can write the decorated node to the JCR FAILED
    java.lang.UnsupportedOperationException at ProtoNodeDecoratorSpec.groovy:136

viveksachdeva:
Please ignore my comment regarding test cases

src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
jbornemann: It was changed due to the message restructure, but the issue @pgoodrich found could occur with the old line as well.

sagarsane: Ok .. but I am skeptical that the issue @pgoodrich found occurs with the old line .. since we have been doing the "ETC" job sync forever without any issues (and the case he mentioned was under /etc/segmentation

jbornemann: Hmm, well either way, now we have coverage over the case; and it's not an issue.

sagarsane:
+1 test for the actual content sync functionality. Looks fine to me. (also, I went ahead and just used brew to install latest protobuf)

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesProcessor.groovy
viveksachdeva: @jbornemann 
```groovy
properties.findResults {JcrProperty jcrProperty ->
isPropertyTransferable(jcrProperty) ? toProperty(jcrProperty):null
}
```

viveksachdeva:
I was having issues with both sever and client on my local. I had author as server and publish as client   but when I do that, startHttpConnection task fails.. :(

Moreover, tried installing new protobuf version on a Linux machine.. Getting error in make task.. The shell script that was added did not work for me...

viveksachdeva:
@jbornemann : So I did one more test with a vanilla AEM as server(different than localhost) and my instance as a client and synched geometrixx website.. Though it may not be related to this PR, but I got following error:

com.day.cq.tagging.impl.TagValidatingEventListener invalid/duplicated tags detected at /content/geometrixx/en/blog/2009/01/dsc_2008_show_inber/jcr:content/cq:tags modified by admin: 'marketing:interest/business, marketing:interest/product' => ''
04.08.2016 11:52:25.404 *ERROR* [pool-5-thread-18] com.day.cq.tagging.impl.TagValidatingEventListener Error while handling event '4'
javax.jcr.ValueFormatException: cq:tags = marketing:interest/business is single-valued.

Moreover dud to some reason, I am not able to use localhost as Grabbit server.. Not sure why.. It fails on startHttpConnection

sagarsane:
@viveksachdeva if you are using our vagrant instances for testing, you cannot use `localhost`. Use `author.twcable.local` or `publish.twcable.local`

jbornemann:
@viveksachdeva you were probably missing some tooling. configure should spit out what's missing. I tried on Centos, and was missing g++ initially. With mac, we get that with XCode. 

Sagar is spot on about your other issue. You have to use our bridge adapter to communicate between VMs (which is mapped to author.twcable.local, publish.twcable.local)

pgoodrich:
@jbornemann Hmm I also started seeing the `cq:tags = business:whatever is single-valued.` error on a page after doing a content sync.  Should check that it's not an issue with grabbit and just a content issue before the next release.

jdigger:
Just to be explicit: The messages are fully compatible between Grabbit v5.0 and Grabbit v6.0. Both v6.0 -> v5.0 and v5.0 -> v6.0, right? Since this is changing a major version of the serialization library and the formal definition was changed, I wanted to make sure this has been properly tested.

jbornemann:
@jdigger they are not compatible with this change. Machines running Grabbit 6 will have to talk to other machines running Grabbit 6

viveksachdeva:
@jbornemann @sagarsane : I tested it with grabbit 5.0.1... I see the same error(but without stacktrace)
05.08.2016 05:42:28.579 *INFO* [pool-5-thread-2] com.day.cq.tagging.impl.TagValidatingEventListener invalid/duplicated tags detected at /content/geometrixx/en/blog/unlisted/pages/about_us/jcr:content/cq:tags modified by admin: 'marketing:interest/business' => '

sagarsane:
That's what I thought. Good to know! Thanks for checking!

jdigger:
I expected as much. https://github.com/TWCable/grabbit/blob/master/docs/AEMSupport.adoc needs to be updated then.

Issue #3, opened 2015-03-10
Handle case where a mixin is sent by server by cannot be added on the client


jazeren1:
Encountered exception is as follows (on the client)

```
02.02.2015 22:08:07.101 [clientJobLauncherTaskExecutor-3] com.twcable.grabbit.client.batch.steps.jcrnodes.JcrNodesWriter *INFO* Encountered invalid mixin type while unmarshalling for Proto values : value {
  stringValue: "mix:versionable"
}

02.02.2015 22:08:07.102 [clientJobLauncherTaskExecutor-3] com.twcable.grabbit.client.batch.steps.jcrnodes.JcrNodesWriter *INFO* Encountered invalid mixin type while unmarshalling for Proto values : value {
  stringValue: "mix:lockable"
}

02.02.2015 22:08:07.103 [clientJobLauncherTaskExecutor-3] com.twcable.grabbit.client.batch.steps.jcrnodes.JcrNodesWriter *ERROR* Exception writing JCR Nodes to current JCR Session : session-admin-334713.  javax.jcr.version.VersionException: Unable to perform operation. Node is checked-in.
	at org.apache.jackrabbit.core.ItemValidator.checkCondition(ItemValidator.java:284)
	at org.apache.jackrabbit.core.ItemValidator.checkModify(ItemValidator.java:248)
```

For starters, may be look at how filevault does it : https://github.com/apache/jackrabbit-filevault/blob/trunk/vault-core/src/main/java/org/apache/jackrabbit/vault/fs/impl/io/ImportInfoImpl.java#L228-L234 (using node.checkIn() )

jdigger:
Suspect this is highly related to GH-96 and GH-65

viveksachdeva:
@jazeren1 @jdigger : Are there some specific steps to reproduce this issue...I tried copying nodes with mixins lockable and versionable and I did not get any error..

jbornemann:
@jazeren1 @jdigger @viveksachdeva I think this has more to do with us not properly handling mix:versionable SCM type features properly. 

My presumption for the above example is that we are attempting to write to a version of a node that is checked-in, thus read-only. https://docs.adobe.com/content/docs/en/spec/jcr/2.0/15_Versioning.html For more info 

In these cases I don't know if we want to sync more history instead, or what. If so, we would have to implement weak/strong links first

sagarsane:
[...]

PR #134, opened 2016-07-25
issue-3 check out node if it is checked in before adding mixin

viveksachdeva:
Issue:
https://github.com/TWCable/grabbit/issues/3

Ready for review.
@jdigger @sagarsane @jbornemann @jazeren1 @jayatitothenew

### DIFF
--- a/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
+++ b/src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
@@ -24,7 +24,11 @@ import groovy.util.logging.Slf4j
 import org.apache.jackrabbit.value.DateValue
 
 import javax.annotation.Nonnull
+<<<<<<< e2f1587401462a04db42959dfda48bd0d1d9dc53
 import javax.annotation.Nullable
+=======
+import javax.jcr.ItemNotFoundException
+>>>>>>> checkoutnode if necessary
 import javax.jcr.Node as JCRNode
 import javax.jcr.Property
 import javax.jcr.Property as JcrProperty
@@ -109,6 +113,7 @@ class JcrNodeDecorator {
         return definition.isMandatory()
     }
 
+<<<<<<< e2f1587401462a04db42959dfda48bd0d1d9dc53
     /**
      * Build node and "only" mandatory child nodes
      */
@@ -155,6 +160,39 @@ class JcrNodeDecorator {
         return null
     }
 
+=======
+    void checkoutIfNecessary(javax.jcr.Node node) throws RepositoryException {
+        try{
+            javax.jcr.Node versionableNode = findVersionableAncestor(node);
+            if (versionableNode != null) {
+                if (!versionableNode.isCheckedOut()) {
+                    versionableNode.checkout();
+                }
+            }
+        }
+        catch (Exception exception){
+            log.warn "Could not checkout node ${node.path}"
+        }
+
+    }
+
+    private javax.jcr.Node findVersionableAncestor(javax.jcr.Node node) throws RepositoryException {
+        if (isVersionable(node)) {
+            return node;
+        }
+        try {
+            node = node.getParent();
+            return findVersionableAncestor(node);
+        } catch (ItemNotFoundException e) {
+            // top-level
+            return null;
+        }
+    }
+
+    private boolean isVersionable(javax.jcr.Node node) throws RepositoryException {
+        return node.isNodeType("mix:versionable");
+    }
+>>>>>>> checkoutnode if necessary
 
     Object asType(Class clazz) {
         if(clazz == JCRNode) {
--- a/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
+++ b/src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
@@ -65,7 +65,7 @@ class ProtoNodeDecorator {
         //Write mixin types first to avoid InvalidConstraintExceptions
         final mixinProperty = getMixinProperty()
         if(mixinProperty) {
-            addMixins(mixinProperty, jcrNode, session)
+            addMixins(mixinProperty, jcrNode)
         }
         //Then add other properties
         writableProperties.each { it.writeToNode(jcrNode) }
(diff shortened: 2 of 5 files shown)


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/134" target="_blank">ReviewNinja

src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
```diff
@@ -91,14 +91,22 @@ class ProtoNodeDecorator {
...
-    private static void addMixins(ProtoPropertyDecorator property, JCRNode node) {
+    private static void addMixins(ProtoPropertyDecorator property, JCRNode node, Session session) {
         property.values.valueList.each { ProtoValue value ->
-            if (node.canAddMixin(value.stringValue)) {
-                node.addMixin(value.stringValue)
-                log.debug "Added mixin ${value.stringValue} for : ${node.name}."
+            try{
+                if(!node.isCheckedOut()){
+                    session.getWorkspace().getVersionManager().checkout(node.path)
+                }
```
sagarsane: Actually, thinking about this more, I am wondering if we should be preventing the case for "node is already checked in" in the first place. 

That is, I think there might be an opportunity to check how the server is serializing the nodes and may be there is an issue with how we are serializing the nodes for some case which is leading to the "node is checked in" error on the client.

viveksachdeva: @sagarsane : I tried finding out a case that would give me this exception but could not find any.. This change was based on what I found in the documentation and checking that things that were working before and after the change. Having a case for this would have helped checking different it from server and client side but 😕

jbornemann:
Hey guys, just wanted to reiterate my comment from the ticket. This issue actually has very little to do with mixins. This issue can occur on setting **_any_** property to a versionable node, when it is checked-in.

Try it out in crxde:

1.) Create a /content/test nt:unstructured node
2.) Add mixin "mix:versionable" 
3.) I used the CityTech Groovy Console, but you can use another tool to do the following:

```groovy
 def versionManager = session.getWorkspace().getVersionManager()
versionManager.checkout('/content/test')
def node = session.getNode('/content/test')
node.setProperty('test', 'test')
session.save()
versionManager.checkin('/content/test')

//def versionManager = session.getWorkspace().getVersionManager()
//versionManager.checkin('/content/test')
//def node = session.getNode('/content/test')
//node.setProperty('test', 'test')
//session.save()
```

Try the uncommented code, and it will work fine; but try the commented code (which sets a property on a checked-in node), and it will fail the same way. 

We need to handle this case for any versionable node. Additionally, what do we do about node verisons? Do we attempt to sync those as well? stored internally under jcr:system/jcr:versionStorage

viveksachdeva:
Added code to checkout Node if necessary.. working code.. would refactor it when other PR is merged

src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
```diff
@@ -93,14 +93,16 @@ class ProtoNodeDecorator {
...
     private static void addMixins(ProtoPropertyDecorator property, JCRNode node) {
         property.values.valueList.each { ProtoValue value ->
+            JCRNodeDecorator jcrNodeDecorator = new JCRNodeDecorator(node)
+            jcrNodeDecorator.checkoutIfNecessary(node);
             if (node.canAddMixin(value.stringValue)) {
                 node.addMixin(value.stringValue)
                 log.debug "Added mixin ${value.stringValue} for : ${node.name}."
-            }
-            else {
+            } else {
```
jbornemann: Style elsewhere is to have following conditions on new-lines.

src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
```diff
@@ -105,6 +106,37 @@ class JCRNodeDecorator {
         return definition.isMandatory()
     }
 
+    void checkoutIfNecessary(javax.jcr.Node node) throws RepositoryException {
```
jbornemann: Why are we passing in a node? This is a decorator. We are delegating to a wrapped node (See line 39)

src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
```diff
@@ -105,6 +106,37 @@ class JCRNodeDecorator {
...
+            node = node.getParent();
+            return findVersionableAncestor(node);
+        } catch (ItemNotFoundException e) {
+            // top-level
+            return null;
+        }
+    }
+
+    private boolean isVersionable(javax.jcr.Node node) throws RepositoryException {
+        return node.isNodeType("mix:versionable");
```
jbornemann: According to the spec : 

```
3.13.2 Versionable Nodes
Under simple versioning, a versionable node must be mix:simpleVersionable. Under full versioning, it must be mix:versionable.
```

so we need to check for either mixin. e.g

```groovy
import static javax.jcr.NodeType.MIX_SIMPLE_VERSIONABLE
import static javax.jcr.NodeType.MIX_VERSIONABLE

private boolean isVersionable() {
  return mixinNodeTypes.any { it == MIX_SIMPLE_VERSIONABLE || it == MIX_VERSIONABLE } 
}
```

src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
jbornemann: This isn't something we should expose. If we are writing to a node, we should capture that within the decorator, and checkout appropriately

Rough pseudo code : 

```groovy
setProperty(name, value, type) {
  modify {
    setProperty(name, value, type
  }
}

modify(Closure<Void> doModify) {
  if(isVersionable()) {
    checkout()
    doModify.call()
    checkin()
  }
  else {
    doModify()
  }
}

File: src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy [hunk 5]
```diff
@@ -105,6 +106,37 @@ class JCRNodeDecorator {
         return definition.isMandatory()
     }
 
+    void checkoutIfNecessary(javax.jcr.Node node) throws RepositoryException {
+        try{
+            javax.jcr.Node versionableNode = findVersionableAncestor(node);
+            if (versionableNode != null) {
+                if (!versionableNode.isCheckedOut()) {
+                    versionableNode.checkout();
```
jbornemann: checkout() and most version related features through Node are depreciated in favor of VersionManager which we can get through the current session's workspace.

```
node.session.workspace.versionManager
```

src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
```diff
@@ -105,6 +106,37 @@ class JCRNodeDecorator {
...
+                }
+            }
+        }
+        catch (Exception exception){
+            log.warn "Could not checkout node ${node.path}"
+        }
+
+    }
+
+    private javax.jcr.Node findVersionableAncestor(javax.jcr.Node node) throws RepositoryException {
```
jbornemann: Not sure I understand why we are doing this. For simplicities sake, I think we should only care about writing to this particular node we are wrapping.

src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
viveksachdeva: @jbornemann : I took a look at Sling's implementation for the same and this code is on same lines.. I still need to refactor it. I am holding off onto that because with protobuf update PR and https://github.com/TWCable/grabbit/pull/135, there would be changes in same classes and if I make changes now, I will have to redo those...

src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
viveksachdeva: If a node is checked in, its connected non-versionable subtree become read-only. So if I need to modify any property of a node in this subtree, I would need to find versionable ancestor and perform checkout..
Moreover, if current node is versionable, this method is as good as just performing checkout on that node and not going to its parent

viveksachdeva:
As it needs changes that are dependent on other PRs, removing review tag for now... Will add it back once other changes are done..

viveksachdeva:
Closing because of merge conflicts..

PR #140, opened 2016-08-13
Upgrade Joda-Time to 2.7 and update MANIFEST to accept [2,3)

sagarsane:
For AEM 6.1 and 6.2 compatibility ..

.. Needs to be tested on both AEM 6.1 and AEM 6.2 .. 

I had to also make a minor change in `grabbit.sh` for CURL to work correctly with AEM 6.2

### DIFF
--- a/grabbit.sh
+++ b/grabbit.sh
@@ -30,7 +30,7 @@ function newGrabbitRequest() {
     echo $SET_YELLOW
     echo "Processing....."
     echo $SET_NO_COLOR
-    curl -s -f -X PUT --data-binary "@$configpath" -u $username:$password -D /tmp/grabbit_headers $client$GRABBIT_JOB > /tmp/grabbit
+    curl -s -f -X PUT --header "Content-Type: text/plain" --data-binary "@$configpath" -u $username:$password -D /tmp/grabbit_headers $client$GRABBIT_JOB > /tmp/grabbit
     RSP_CODE=$?
     if [ $RSP_CODE -ne 0 ]; then
         clear
--- a/build.gradle
+++ b/build.gradle
@@ -63,6 +63,10 @@ ext.providerLink = "https://github.com/TWCable/grabbit"
         attributes 'Implementation-Title': project.name
         attributes 'Implementation-Version': project.version
         attributes 'Implementation-Vendor': project.vendor
+
+        instruction 'Import-Package', 'org.joda.time; version="[2,3)"'
+        instruction 'Import-Package', 'org.joda.time.base; version="[2,3)"'
+        instruction 'Import-Package', 'org.joda.time.format; version="[2,3)"'
     }
 
     afterEvaluate {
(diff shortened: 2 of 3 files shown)


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/140" target="_blank">ReviewNinja

grabbit.sh
```diff
@@ -30,7 +30,7 @@ function newGrabbitRequest() {
     echo $SET_YELLOW
     echo "Processing....."
     echo $SET_NO_COLOR
-    curl -s -f -X PUT --data-binary "@$configpath" -u $username:$password -D /tmp/grabbit_headers $client$GRABBIT_JOB > /tmp/grabbit
+    curl -s -f -X PUT --header "Content-Type: text/plain" --data-binary "@$configpath" -u $username:$password -D /tmp/grabbit_headers $client$GRABBIT_JOB > /tmp/grabbit
```
sagarsane: This seems necessary to correctly `PUT` the config file to AEM 6.2 correctly. Without this, in AEM 6.2... the content of the file gets wrongly placed into weirdly managed Request Parameters instead.

This continues to work as it is on AEM 6.1.

sagarsane:
This is ready for testing and review @jdigger @jbornemann @viveksachdeva 

cc @masroormohammed @jazeren1 

The best way to test this would be to use a AEM 6.1 author as server and AEM 6.2 as client and try to copy content from AEM 6.1 author to AEM 6.2 publish.

grabbit.sh
jbornemann: weird..

gradle.properties
```diff
@@ -5,7 +5,7 @@ org.gradle.jvmargs=-XX:MaxPermSize=512m -XX:+CMSClassUnloadingEnabled -XX:+CMSPe
 bundleInstallRoot = /apps/grabbit/install
 
 group = com.twcable.grabbit
-version = 6.0.0
```
jbornemann: Should we keep this as a seperate commit? Makes a bit easier to view RC

sagarsane: Oh yeah for sure. I normally just have it for testing. Will be removing it from this commit before TM.

jbornemann: Oh, gotcha

grabbit.sh
jbornemann: We may need to do something similar for https://github.com/TWCable/grabbit-cli/blob/master/src/main/java/com/twcable/grabbit/tools/jobstarter/JobStarter.java#L148 for 6.2 support

sagarsane: Yeah true .. or at least test it (may be `application/json` will work there)

viveksachdeva:
@sagarsane : Tested with 6.2 author to 6.1 author.. Worked alright...

jbornemann:
Looks good on 6.1 author and pub +1

sagarsane:
Actually, I think this should also be tested on 6.2 publisher

sagarsane:
Tested fine on AEM 6.1 -> AEM 6.2 publisher as well.

jbornemann:
+1

Issue #141, opened 2016-08-14
Multi Valued property with only 1 value in it on Grabbit Server gets written as a Single Valued property on the Grabbit Client


sagarsane:
This seems to be affecting at least the 6.0.0 release.

One example to test this is `cq:tags`.. Try to sync a DAM Asset  which has only 1 tag assigned to it ..

Grabbit Server: 

![grabbitservercqtags](https://cloud.githubusercontent.com/assets/667114/17647132/4e87c08a-61b3-11e6-8767-5615c4a9deb2.jpg)

Grabbit Client:

![grabbitclientcqtags](https://cloud.githubusercontent.com/assets/667114/17647137/763d63a0-61b3-11e6-93f1-021d4b5bfef7.jpg)

Above has an example of how Grabbit Client is writing when Grabbit Server has a multivalued property with just single value in it.

sagarsane:
Fixed by #142

PR #142, opened 2016-08-14
Transfer current property's state whether its multiple or not

sagarsane:
Possible solution for #141 

Ofcourse. there are a lot of tests failing for this currently

@jdigger @jbornemann what do you think of this approach? Other approach I thought of was to go back and add the `value` message that was removed in a recently ..

### DIFF
--- a/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
+++ b/src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
@@ -82,16 +82,6 @@ class ProtoPropertyDecorator {
         innerProtoProperty.name == JCR_MIXINTYPES
     }
 
-
-    boolean isMultiple() {
-        /**
-         * According to the JCR spec, all properties must have a value; however, a multi-value property can have an "empty" value (like an empty array)
-         * That is how we know it is a multiple valued property if the values count is 0
-         */
-        return valuesCount > 1 || valuesCount == 0
-    }
-
-
     ProtoValue getValue() {
         innerProtoProperty.valuesList.first()
     }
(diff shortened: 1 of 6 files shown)


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/142" target="_blank">ReviewNinja

jbornemann:
Yeah, I think this is a good way to go 💭

sagarsane:
I've fixed the tests. This is ready for review and testing

@jbornemann @viveksachdeva @jdigger @masroormohammed @jazeren1

src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
```diff
@@ -82,20 +82,6 @@ class ProtoPropertyDecorator {
         innerProtoProperty.name == JCR_MIXINTYPES
     }
 
-
-/*
-    boolean isMultiple() {
```
viveksachdeva: this method is used at a few places in this class... wont that be an issue?

sagarsane: It will now use the `isMultiple()` method from the generated Java class from the updated node.proto message

viveksachdeva:
With 6.0.0, single valued property is created and getting following error
19.08.2016 05:28:17.785 *INFO* [pool-5-thread-11] com.adobe.cq.creativecloud.filesync.impl.events.SyncEventGenerator Asset added at /content/dam/abcjpg.
19.08.2016 05:28:17.793 *ERROR* [pool-5-thread-9] com.day.cq.tagging.impl.TagValidatingEventListener Error while handling event '4'
javax.jcr.ValueFormatException: cq:tags = ccc is single-valued.
      
With this branch, array type property with one value is created and not seeing above mentioned error..

+1 Testing

Issue #65, opened 2015-09-01
Protected user/group properties


jdigger:
Running against `/home` complains about mandatory property `"rep:principalName"` being missing.

Users/groups are being stored in Authorizable nodes. Even though both sides are running as "admin", `theSession().save()` in `JcrNodesWriter` gives errors like: `javax.jcr.nodetype.ConstraintViolationException: OakConstraint0021: /home/groups/b[[rep:Group]]: Mandatory property rep:principalName not found in a new node`

The issue is that this is a protected property in CRX 2.4 / JCR so the JCR API can not be used. (c.f. [Interaction With Protected Properties](http://www.day.com/specs/jcr/2.0/16_Access_Control_Management.html#16.3.12%20Interaction%20with%20Protected%20Properties), [Repository Model - Protected](http://www.day.com/specs/jcr/2.0/3_Repository_Model.html#3.7.2.2%20Protected).)

Before writing we need to check if the node type to be written is a user/group and use [Jackrabbit's UserManager API](http://docs.adobe.com/docs/de/aem/6-0/develop/ref/javadoc/org/apache/jackrabbit/api/security/user/UserManager.html) to create the node.
(see `com.twcable.grabbit.client.batch.steps.jcrnodes.JcrNodesWriter` in Grabbit.)

dzmitry-kankalovich:
Just was passing by this issue. Could be the reason: https://blogs.perficientdigital.com/2019/01/15/including-service-users-in-your-aem-package/

PR #143, opened 2016-08-16
issue-3-handle-mixin-property

viveksachdeva:
Issue : https://github.com/TWCable/grabbit/issues/3

WIP


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/143" target="_blank">ReviewNinja

src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy
```diff
@@ -25,6 +25,7 @@ import org.apache.jackrabbit.value.DateValue
```
jbornemann: Name of this file is different from the class. JcrNodeDecorator/JCRNodeDecorator

src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
```diff
@@ -97,6 +97,10 @@ class ProtoNodeDecorator {
                 log.debug "Added mixin ${value.stringValue} for : ${node.name}."
             }
             else {
+                JcrNodeDecorator decoratedNode = new JcrNodeDecorator(node)
```
jbornemann: I think these are details that ProtoNodeDecorator, or any other consumer probably shouldn't know about. ProtoNodeDecorator just wants to write a mixin in this case; It shouldn't need to know the details. 

To extend that thought, I think we are doing too much within writeToJcr. We should hand JcrNodeDecorator a ProtoNodeDecorator, and it should have all the information needed to build itself.

src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
```diff
@@ -97,6 +97,10 @@ class ProtoNodeDecorator {
...
+                if(decoratedNode.checkoutIfNecessary()){
+                    addMixins(property, node)
```
jbornemann: To reiterate, this isn't just about writing mixins. If a node is checked in, we can't write/modify any property, or child node within or under that node. 

I think within JcrNodeDecorator, we need to wrap things like setProperty, and addMixin; intercept these calls; THEN perform checkout/checkin actions as needed

src/main/groovy/com/twcable/grabbit/jcr/ProtoPropertyDecorator.groovy
```diff
@@ -70,6 +71,12 @@ class ProtoPropertyDecorator {
                 log.warn "WARNING!  Property ${name} will not be written to ${node.name}!  There was a problem when writing value type ${PropertyType.nameFromValue(type)}${multiple ? '[]' : ''} to existing node with same type, due to a ValueFormatException, and we were unable to recover"
             }
         }
+        catch (VersionException ex){
```
jbornemann: We need to be proactive and check for this case up front; especially since this isn't an exception case. Intercept, and act. See some of my comments on your previous PR.

viveksachdeva:
Need to fix existing test case for JcrNodeWriterSpec and add a new one for writeToJcr in JcrNodeDecoratorSpec.... Still trying to figure that out..

src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
```diff
@@ -88,8 +88,9 @@ class JcrNodesWriter implements ItemWriter<ProtoNode>, ItemWriteListener {
     }
 
     private static void writeToJcr(ProtoNode nodeProto, Session session) {
```
jbornemann: All this logic needs to be moved within ProtoNodeDecorator, the above "void write()" will be : 

```groovy
void write(List<? extends ProtoNode> nodeProtos) throws Exception {
  Session session = theSession()
  for(ProtoNode node : nodeProtos) {
    new ProtoNodeDecorator(node).writeToJcr(session)
  }
}

File: src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy [hunk 6]
```diff
@@ -88,8 +88,9 @@ class JcrNodesWriter implements ItemWriter<ProtoNode>, ItemWriteListener {
...
-        JcrNodeDecorator jcrNode = new ProtoNodeDecorator(nodeProto).writeToJcr(session)
-        jcrNode.setLastModified()
+        ProtoNodeDecorator protoNodeDecorator = new ProtoNodeDecorator(nodeProto)
+        JcrNodeDecorator nodeDecorator = new JcrNodeDecorator(protoNodeDecorator.getOrCreateNode(session))
+        nodeDecorator.writeToJcr(protoNodeDecorator, session)
         // This will processed all mandatory child nodes only
         if(nodeProto.mandatoryChildNodeList && nodeProto.mandatoryChildNodeList.size() > 0) {
```
jbornemann: We are trying to keep application logic consolidated within our application objects (ProtoNodeDecorator, JcrNodeDecorator), and out of the framework (JcrNodesWriter). One thing can change without the other needing to know. In addition, we can use these objects in multiple places within the framework without worrying about the details.

src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
```diff
@@ -59,21 +59,6 @@ class ProtoNodeDecorator {
         protoProperties.findAll { !(it.name in [JCR_PRIMARYTYPE, JCR_MIXINTYPES]) }
     }
 
-
-    JcrNodeDecorator writeToJcr(@Nonnull Session session) {
```
jbornemann: Why is this deleted?

src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy
```diff
@@ -135,6 +139,66 @@ class JcrNodeDecorator {
...
+    private boolean isVersionable() throws RepositoryException {
+        return mixinNodeTypes.any { it.name in ["mix:simpleVersionable", MIX_VERSIONABLE] }
+    }
+
+    /*
+    * Write proto properties to Jcr Node
+    * Mixins are added before other properties
+    */
+
+    void writeToJcr(ProtoNodeDecorator decoratedProtoNode, @Nonnull Session session) {
```
jbornemann: I'm confused. This is a JcrNodeDecorator. It wraps a node that is already within the JCR. Why would we call writeToJcr on an instance of this object and pass a ProtoNodeDecorator?

viveksachdeva: @jbornemann : This method is writing mixin and other properties to the Jcr node... Shouldnt this operation be in one of the two wrapper objects : JcrNodeDecorator or JcrPropertyDecorator

src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
viveksachdeva: From what I thought was that all the JCR operations should be under the Jcr decorator classes and Proto Decorators should be used for proto related operations only.. In that case, moving this to ProtoNodeDecorator would mean creating a JcrNodeDecorator there and passing necessary values to jcr decorator methods..

src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
viveksachdeva: This is what I was trying to say in other comment.. ProtoNodeDecorator to be passed to JcrDecorator and that should be doing write part

src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy
jbornemann: This is getting confusing. 

A ProtoNodeDecorator wraps a ProtoNode, and provides inspections on the ProtoNode. In addition, it knows how to write it self to the JCR.

A JcrNodeDecorator wraps a node already within the JCR, and provides inspections on that node. In addition, it knows how to write it self to a ProtoNode...

If now we have a multi-step process for creating a JcrNode from a Protonode (like making sure an ancestor is checked out before saving the session, and making the node concrete), maybe that means we need something else. Steps being, create the node and add properties, see if ancestor needs to be checked out, checkout if needed, persist (save session). 

```groovy
JcrNodeDecorator writeToJcr(@Nonull Session session) {
  final node = getNode(session)
  final mixinProperty = getMixinProperty()
  if(mixinProperty) {
    transientNode.addMixins(mixinProperty)
  }
  writableProperties.each { transientNode.writeProperty(it) }
  return transientNode.save()
} 

JcrNodeDecorator getNode(session) {
  JcrNodeDecorator.from(JcrUtils.getOrCreateByPath(......))
}

//Previous concrete implementation goes in here :
class ConcreteJcrNodeDecorator implements JcrNodeDecorator {}

class TransientJcrNodeDecorator implements JcrNodeDecorator {

  Node unsavedNode

  TransientJcrNodeDecorator(JCRNode node) {
    if (!node.isNew()) throw new IllegalStateException(...)...
   }

  ConcreteJcrNodeDecorator save() {
    doSave {
     unsavedNode.save()
   }
  }

  void doSave(Closure<Void> doSave) {
    checkoutIfNeeded()
    doSave.call()
    checkinIfNeeded()
  }

  void checkoutIfNeeded() {
    //check versionable, or ancestor versionable
    //if so, checkout
  }
  void writeProperty(...) {}
   ...
  ...
}

Comment (jbornemann) [inline src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy]: Just Pseudocode, but hopefully it is followable

File: src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy (hunk 6 above)
Comment (jbornemann) [inline src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy]: _"From what I thought was that all the JCR operations should be under the Jcr decorator classes and Proto Decorators should be used for proto related operations only"_ 

This will probably be mostly true. What I'm referencing above doesn't conform to this. We have a third actor creating decorators; passing decorators to decorators; having some decorators write their own nodes; having third actor inspecting decorators, and writing nodes elsewhere...

The point is to have this logic in a place where it is cohesive. In other words, it tends to change with the things that change around it.

File: src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy (hunk 8 above)
Comment (jbornemann) [inline src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy]: Not necessarily...the design can evolve..

See below

File: src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy (hunk 7 above)
Comment (jbornemann) [inline src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy]: Actually the above doesn't make sense...

**Comment by viveksachdeva** [comment]
@jbornemann : NOT ready for review.. Would discuss my changes and then mark it for review.. :)

File: src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy (hunk 1 above)
Comment (viveksachdeva) [inline src/main/groovy/com/twcable/grabbit/jcr/JCRNodeDecorator.groovy]: I have now changed the file name but it is not getting reflected here... :(

File: src/main/groovy/com/twcable/grabbit/jcr/ProtoNodeDecorator.groovy [hunk 9]
```diff
@@ -59,9 +59,25 @@ class ProtoNodeDecorator {
...
+        }
+    }
 
-    JcrNodeDecorator writeToJcr(@Nonnull Session session) {
+
+    private JcrNodeDecorator writeToJcr(@Nonnull Session session) {
         final jcrNode = getOrCreateNode(session)
+
+        JcrNodeDecorator decoratedJcrNode = new JcrNodeDecorator(jcrNode)
+        if (!decoratedJcrNode.isCheckedOut()) {
```
viveksachdeva: @sagarsane , @jbornemann : As we create a JcrNodeDecorator object here, I think it makes sense to do checkout thing in this method... Thoughts?

src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy
```diff
@@ -136,6 +136,41 @@ class JcrNodeDecorator {
...
+        try {
+            JcrNodeDecorator parentDecoratedNode = new JcrNodeDecorator(this.parent);
+            return parentDecoratedNode.findVersionableAncestor();
+        } catch (ItemNotFoundException e) {
+            return null;
+        }
+    }
+
+    /**
+     * mix:simpleVersionable cannot be checked in. Throws javax.jcr.UnsupportedRepositoryOperationException
```
jbornemann: I'd say add both. We idol worship the JSR-283 before Jackrabbit 🙏 

If Jackrabbit decides to be more compliant, then or application won't break.

src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy
```diff
@@ -136,6 +136,41 @@ class JcrNodeDecorator {
...
+     * Checking here for mix:versionable only
+     * */
+    private boolean isVersionable() throws RepositoryException {
```
jbornemann: Minor, but I would remove the RepositoryException declaration here. Almost all JCR operations <i>potentially</i> through this exception as a catch all "something went wrong" in the vaguest of ways.   From our perspective, it's just sort of noise - We can't recover, and we can't really learn directly from it.

src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy
```diff
@@ -136,6 +136,37 @@ class JcrNodeDecorator {
...
+        try {
+            JcrNodeDecorator parentDecoratedNode = new JcrNodeDecorator(this.parent)
+            return parentDecoratedNode.findVersionableAncestor()
+        } catch (ItemNotFoundException e) {
+            return null
+        }
+    }
+
+    private boolean isVersionable() {
+        return isNodeType(MIX_VERSIONABLE) || isNodeType("mix:simpleVersionable")
```
viveksachdeva: Not using NodeType class as it has 
MIX_SIMPLE_VERSIONABLE = "{http://www.jcp.org/jcr/mix/1.0}simpleVersionable";
and wanted to check for mix:simpleVersionable

jbornemann: isNodeType(String) works with the FQN of node type, or shorthand (e.g nt:unstructured). 

NodeType.MIX_VERSIONABLE actually equals the longhand "{http://www.jcp.org/jcr/mix/1.0}versionable"

viveksachdeva: Yeah.. so I did not use NodeType class to refer simple version... the one that you mentioned is referred from jcr constant that is shor value

jbornemann:
@viveksachdeva  take a look at https://github.com/TWCable/grabbit/compare/e4d3656...issue-3-jeff 

These were some of my ideas around how we could organize some of this stuff. I figured it would be easier to write code down, than attempt to explain all of it, but the jist : 

1.) Trying to keep operations on data, close to where the data is organized. IMO, in general, if you see something like ProtoPropertyDecorator being passed into a method within JcrNodeDecorator, it probably means that whatever operations are being performed on ProtoPropertyDecorator should happen within ProtoPropertyDecorator.

2.) Since we do have to potentially search up the tree to find the root versionable node; For performance, it probably makes sense to have this be the exceptional case, rather than the norm. Scaling up the tree for deeply-nested, unversionable nodes would be expensive.

3.) If we find a versionable node, and check it out, we should check it back in when we are done with it.

src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
```diff
@@ -71,7 +70,7 @@ class JcrNodesWriter implements ItemWriter<ProtoNode>, ItemWriteListener {
     void write(List<? extends ProtoNode> nodeProtos) throws Exception {
         Session session = theSession()
         for (ProtoNode nodeProto : nodeProtos) {
-            writeToJcr(nodeProto, session)
+            new ProtoNodeDecorator(nodeProto).writeJcrNodesRecursively(session)
```
jbornemann: Too much detail in the method name. From JcrNodesWriter's perspective (or any client), we only care that our ProtoNodeDecorator is written correctly as a JcrNodeDecorator. Whatever it takes to get to that point - details like writing mandatory child nodes recursively, and checking out versionables should be kept out of sight.

src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy
```diff
@@ -136,6 +136,37 @@ class JcrNodeDecorator {
...
+     * Finds out nearest versionable ancestor for a node and performs a checkout
+    */
+    void checkoutNode() {
+        try {
+            JcrNodeDecorator decoratedVersionableAncestor = findVersionableAncestor()
+            if (decoratedVersionableAncestor && !decoratedVersionableAncestor.isCheckedOut()) {
+                decoratedVersionableAncestor.session.workspace.versionManager.checkout(decoratedVersionableAncestor.path)
+            }
+        }
+        catch (Exception exception) {
```
jbornemann: Log particular exceptions

src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy
jbornemann: Ok, well should we use javax.jcr.nodetype.NodeType?...

I'd rather us reference constants, then manage that stuff

src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
sagarsane: Also, assigning a variable here would make it easier to debug IMO!

```groovy
            ProtoNodeDecorator protoNodeDecorator = new ProtoNodeDecorator(nodeProto)
            protoNodeDecorator.writeToJcr(session)
```

sagarsane:
@viveksachdeva .. playing catchup here.. but can you provide steps to reproduce the case that breaks without this change (of explicitly checking out the node and writing mixin)?

sagarsane:
Nvm @viveksachdeva .. @jbornemann pointed me to https://github.com/TWCable/grabbit/pull/134#issuecomment-236684066 . Just trying to reproduce the original issue.

viveksachdeva:
@jbornemann : When refactoring JcrProcessor, per the discussion that we had, ProtoNodeDdecorator should know how to write itself to Jcr node and JcrNodeDecorator would have JcrNode wrappers and ability to build proto node from it but as per https://github.com/TWCable/grabbit/compare/e4d3656...issue-3-jeff#diff-0e12e515926b4dbd7a7b92edebc8f7ecR66, we are delegating creating jcr node from proto to JcrNodeDecorator..

src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy
viveksachdeva: It could throw UnsupportedRepositoryOperationException, LockException or RepositoryException.. the log statement includes the exception message... IMO if we are not doing any different for different types of exceptions, having single exception is cleaner...

src/main/groovy/com/twcable/grabbit/client/batch/steps/jcrnodes/JcrNodesWriter.groovy
viveksachdeva: agreed.. I wanted to differentiate this method from other write to jcr and this found this name relevant to what the method is doing... will change it to something more appropriate

src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy
sagarsane: So luckily .. first 2 extend `RepositoryException` .. so you'd just need to do catch `RepositoryException` .. in-fact IMO if we get any exception that is other than a type of `RepositoryException` ... we should break.

viveksachdeva:
@jbornemann : Regarding 1) writeToNode is already there in ProtoPropertyDecorator.. is there something else that you are referring to in this comment.

2,3) If for checkout, it traverses till the root node, it would happen only once. From the next time, isCheckedOut() would return true and code for findingVersionableAncestor would not run again... IMO we do a check early for checkout as if that fails, we would anyway be getting the exception and then we would checkout the node and run same code again.. 
Moreover if we do it in exception block https://github.com/TWCable/grabbit/compare/e4d3656...issue-3-jeff#diff-412149750a01ec3f3187ff7ce351cd17R65, we would need to do it for mixins and other properties separately.. So I think this check should be done before we start to write anything in the node..
And related to node checkin, I think checkin should not be needed as we will be creating version of the new content may not be required... We have anyway modified what was read only, why set it back to same status..

jbornemann:
1.) Yes. It's not about writeToNode being in ProtoNodeDecorator. ProtoNodeDecorator.writeToJcr should be there to gain a handle on a JcrNodeDecorator. What it should not do is have details on how JcrNodeDecorator is created....that should happen within JcrNodeDecorator. We are trying to create highly cohesive, SRP objects. Merge my commit, and it will be more clear - my commit reflects the direction I think we should be going in. It will be even more clear why this design direction is a good idea as my PR for users/groups goes in. 

2/3) If we checkout a node, we should check it back in. Leaving all of the client's doors open is not acceptable. Furthermore, "If for checkout, it traverses till the root node, it would happen only once" - This is not true, because my previous point, and second we can not have any assumptions on the state of which a node is in - for all we know, the application is changing it all of the time. Every node we sync will be scaling up to either the root (if it is unversionable), or to a common versionable root. If we have to perform this behavior, again, lets only do it when we have to.

jbornemann:
"ProtoNodeDdecorator should know how to write itself to Jcr node and JcrNodeDecorator would have JcrNode wrappers and ability to build proto node from it but as per e4d3656...issue-3-jeff#diff-0e12e515926b4dbd7a7b92edebc8f7ecR66, we are delegating creating jcr node from proto to JcrNodeDecorator.."

What I mean, is from the client perspective, ProtoNodeDecorator should have methods that allow me, the client, to get a JcrNodeDecorator from a ProtoNodeDecorator. As I client, I don't care how it happens. As the writer of ProtoNodeDecorator, I care how it happens, for the same reasons. 

Think about it - if the details of building a JcrNodeDecorator change, where do those changes need to occur in the application?  If I have ProtoNodeDecorator (or any other client! JcrProcessor even) setting state, checking for state (e.g hasMixins) in order to build a JcrNodeDecorator correctly; than that means whenever representation of state changes for JcrNodeDecorator, I have to change all of the objects that act on that state. If JcrNodeDecorator builds itself by setting, and checking internal state; I only have to change JcrNodeDecorator. 

The why we do this is we only want to give an object one reason to change, and therefore one reason to break - itself. We don't want objects breaking because other objects have changed, or have broken.  This is the Single Responsibility Principle.

src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy
jbornemann: The problem with catching "Exception", and in some cases other broad unchecked exceptions is that they bury programming errors, along with run time errors - which is bad. We want to know if we are throwing NullPointerExceptions. We want the application to break. We want the tests to fail. We don't just want log messages in those cases.

src/main/groovy/com/twcable/grabbit/jcr/JcrNodeDecorator.groovy
```diff
@@ -45,11 +49,41 @@ class JcrNodeDecorator {
...
 
+    static JcrNodeDecorator createFromProtoNode(@Nonnull final ProtoNodeDecorator protoNode, @Nonnull final Session session) {
+        final JcrNodeDecorator theDecorator = new JcrNodeDecorator(getOrCreateNode(protoNode, session))
+        final mandatoryNodes = protoNode.getMandatoryChildNodes()
+        mandatoryNodes?.each {
+            it.writeToJcr(session)
+        }
+        //If a version exception is thrown,
+        try {
+            protoNode.writableProperties.each { it.writeToNode(innerNode) }
```
viveksachdeva: @jbornemann : I have merged your commit to this... Now we are adding mixins along with other properties whereas earlier we were adding mixins first and then other props... and there was a comment for the same:
-        //Write mixin types first to avoid InvalidConstraintExceptions
Just wanted to confirm is this will not be an issue before changing it again

viveksachdeva:
Most of this taken care by @jbornemann in #65 .. Will create a new one in case something is left..

Issue #91, opened 2015-11-10
Add an ability to give order the paths in Grabbit Configs


sagarsane:
There should be some way of ordering paths in a grabbit config. 

This could be manual/user driven (something like `order:1`, `order:2` for each path in the config) or automated / programmatically calculated in some way (by looking at the paths in question and figuring out which paths should go in first)

jdigger:
This is to address the sequencing issue caused by concurrency and the issue described in GH-92.

jbornemann:
A potential solution pasted from https://github.com/TWCable/grabbit/pull/144: 

1.) We need to create a ranking for paths. Here are some example paths for demonstration:

/foo - 1
/foo/bar - 2
/foo/bar/zoo - 3
/foo/car/do - 1

Start each path out with a ranking of 1, then for each path in sorted order based on path depth, increment the path's ranking by it's parent's ranking if it is found within the visited set; otherwise keep a ranking of 1.

This makes sure paths that can be processed independently are within the same rank group.

After that, we need to make sure Spring Batch will feed the jobs from our job queue, into the thread pool in order, based on ranking. To do this, we would need a Spring ThreadPoolTaskExecutor instance with a ThreadPoolExecutor that is built with a BlockingPriorityQueue implemented with a Comparator that compares our jobs based on path ranking. Long winded, I know..

Still even then, this would only make sure that jobs are feed into the pool in priority order, but it wouldn't prevent certain sized thread pools from picking up and starting work on higher ranked jobs before lower ranked jobs complete. To solve this, we would probably need some sort of system within our validator task that polls the job queue to see if lower ranked jobs are still in flight, and if so, put the thread of the higher ranked job to sleep for some period.

Shashi-Bhushan:
Had a discussion with @jbornemann regarding this, quoting him from the discussion

> Any solution will be too complicated compared to the value it would give us. At the very minimum, what we have now (checking to see if path already has an existing parent before running the path's job - and raising a configuration error) I think is acceptable.

Closed the PR i created couple of weeks back. Should this ticket be closed as well, or we keep it in To-Do list for a later date.
@jdigger @jazeren1 @sagarsane @viveksachdeva

sagarsane:
Thanks @Shashi-Bhushan .. Yes I think we should keep this open for now.

PR #144, opened 2016-08-17
Path order specified in Groovy

Shashi-Bhushan:
WIP for issue #91

### DIFF
--- a/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
+++ b/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
@@ -107,7 +107,7 @@ class GrabbitConfiguration {
 
         if (errorBuilder.hasErrors()) throw errorBuilder.build()
 
-        sortPathConfigByDepth(pathConfigurations)
+        pathConfigurations = sortPathConfigs(pathConfigurations)
 
         return new GrabbitConfiguration(
             serverUsername,
@@ -120,15 +120,26 @@ class GrabbitConfiguration {
     }
 
     /**
-     * Sorts the {@code pathConfigurations} on the basis of path depth
+     * Removes '/.' from {@code pathConfigurations}'s path if exists in the end, and Sorts the {@code pathConfigurations}
+     * lexicographically on the basis of paths
      *
      * @param pathConfigurations
      *          Collection of {@link PathConfiguration} to sort
      * @return
      *          Sorted Collection of {@link PathConfiguration}
      */
-    private static Collection<PathConfiguration> sortPathConfigByDepth(Collection<PathConfiguration> pathConfigurations){
-        pathConfigurations.sort{
+    private
+    static Collection<PathConfiguration> sortPathConfigs(@Nonnull final Collection<PathConfiguration> pathConfigurations){
+        Collection<PathConfiguration> modifiablePathConfig = pathConfigurations;
+
+        modifiablePathConfig.collect{ element ->
+            if(element.path.substring(element.path.lastIndexOf('/')) == '/.') {
+                element.path = element.path.substring(0, element.path.lastIndexOf('/'))
+            }
+            return element
+        }
+
+        modifiablePathConfig.sort{
             it.path
         }
     }
--- a/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
+++ b/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
@@ -120,22 +120,21 @@ class GrabbitConfiguration {
     }
 
     /**
-     * Removes '/.' from {@code pathConfigurations}'s path if exists in the end, and Sorts the {@code pathConfigurations}
-     * on the basis of paths' depth
+     * Sorts the {@code pathConfigurations} on the basis of paths' depth
      *
      * @param pathConfigurations
      *          Collection of {@link PathConfiguration} to sort
      * @return
-     *          Sorted Collection of {@link PathConfiguration}
+     *          Sorted List of {@link PathConfiguration}
      */
-    private static Collection<PathConfiguration> getSortedPathConfigurationsByDepth(
+    private static List<PathConfiguration> getSortedPathConfigurationsByDepth(
             @Nonnull final Collection<PathConfiguration> pathConfigurations) {
 
-        return pathConfigurations.sort(false) { PathConfiguration it ->
-            final String thePath = it.path
+        return pathConfigurations.sort(false) { PathConfiguration pathConfig ->
+            String thePath = pathConfig.path
             //'/.' means 'this node', so we don't count it as an additional directory level
-            it.path = thePath[-2..-1] == "/." ? thePath - "/." : thePath
-            return it.path.count("/")
+            thePath = thePath[-2..-1] == "/." ? thePath - "/." : thePath
+            return thePath.count("/")
         }
     }
 
--- a/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
+++ b/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
@@ -107,7 +107,7 @@ class GrabbitConfiguration {
 
         if (errorBuilder.hasErrors()) throw errorBuilder.build()
 
-        sortPathConfigByPath(pathConfigurations)
+        sortPathConfigByDepth(pathConfigurations)
 
         return new GrabbitConfiguration(
             serverUsername,
@@ -120,14 +120,14 @@ class GrabbitConfiguration {
     }
 
     /**
-     * Sorts the {@code pathConfigurations} on the basis of path in configuration
+     * Sorts the {@code pathConfigurations} on the basis of path depth
      *
      * @param pathConfigurations
      *          Collection of {@link PathConfiguration} to sort
      * @return
      *          Sorted Collection of {@link PathConfiguration}
      */
-    private static Collection<PathConfiguration> sortPathConfigByPath(Collection<PathConfiguration> pathConfigurations){
+    private static Collection<PathConfiguration> sortPathConfigByDepth(Collection<PathConfiguration> pathConfigurations){
         pathConfigurations.sort{
             it.path
         }
--- a/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
+++ b/src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
@@ -128,18 +128,18 @@ class GrabbitConfiguration {
      * @return
      *          Sorted Collection of {@link PathConfiguration}
      */
-    private
-    static Collection<PathConfiguration> sortPathConfigs(@Nonnull final Collection<PathConfiguration> pathConfigurations){
+    private static Collection<PathConfiguration> sortPathConfigs(
+            @Nonnull final Collection<PathConfiguration> pathConfigurations) {
         Collection<PathConfiguration> modifiablePathConfig = pathConfigurations;
 
-        modifiablePathConfig.collect{ element ->
-            if(element.path.substring(element.path.lastIndexOf('/')) == '/.') {
+        modifiablePathConfig.collect { element ->
+            if (element.path.substring(element.path.lastIndexOf('/')) == '/.') {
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/144" target="_blank">ReviewNinja

Shashi-Bhushan:
[Issue Link](https://github.com/TWCable/grabbit/issues/91): Sorting path configurations on the basis of path depth.

- [x] Sort pathConfigurations by path depth


@sagarsane @jbornemann @jdigger @viveksachdeva 
Please have a look at this one.

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -117,6 +119,20 @@ class GrabbitConfiguration {
         )
     }
 
+    /**
+     * Sorts the {@code pathConfigurations} on the basis of path depth
```
jbornemann: "Sorts the .... on the basis of path depth". 

That's not really what we are doing below. We are sorting paths within the collection lexicographically.

What about an obscure case where we have two paths, "/foo/." - for creating just the root node, and "/foo/$" for creating a node named "$", under "/foo".  "/foo/$" would come first lexicographically. 

I think we need to strip the leading node (or '.'), and forward slash; then compare by counting the number of forward slashes within paths.

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -117,6 +119,20 @@ class GrabbitConfiguration {
...
+     *
+     * @param pathConfigurations
+     *          Collection of {@link PathConfiguration} to sort
+     * @return
+     *          Sorted Collection of {@link PathConfiguration}
+     */
+    private static Collection<PathConfiguration> sortPathConfigByDepth(Collection<PathConfiguration> pathConfigurations){
```
jbornemann: Generally not a good idea to mutate an argument like this. In addition, we are mutating the argument, returning it; then throwing away the return above.

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
Shashi-Bhushan: Hello @jbornemann 
for handling this case, I've removed '/.' from the end before sorting the pathConfigurations lexicographically. Added a spec for clarification as well.
Please suggest if this would work, and also, if there are other edge cases may be, that needs to be taken care of.

Thanks

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -117,6 +119,31 @@ class GrabbitConfiguration {
...
+     * @param pathConfigurations
+     *          Collection of {@link PathConfiguration} to sort
+     * @return
+     *          Sorted Collection of {@link PathConfiguration}
+     */
+    private static Collection<PathConfiguration> sortPathConfigs(
+            @Nonnull final Collection<PathConfiguration> pathConfigurations) {
+        Collection<PathConfiguration> modifiablePathConfig = pathConfigurations;
+
+        modifiablePathConfig.collect { element ->
```
Shashi-Bhushan: @jazeren1 @jdigger @viveksachdeva @jbornemann 
i just noticed that the original pathConfiguration's path is also changing here. i think it's because collect works on a shallow copy of modifiablePathConfig.
though i'm using collecting the returned value of the method now, still mutation of the original one before that re-assignment does not seem right somehow.

your thoughts please ?

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
jbornemann: I'm still of the opinion that we should implement this by :

```
I think we need to strip the leading node (or '.'), and forward slash; then compare by counting the number of forward slashes within paths.
```

I the way it's implemented now may accomplish the goal of having a given parent path before child path(s), but I may not be smart enough to think of all the ways it could break.  Additionally, it may be misconstrued in the future for doing something it's not - like sorting based on path depth. In my opinion, counting directory levels is a bit more straight forward.

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
jbornemann:
```groovy
private static List<PathConfiguration> getSortedPathConfigurationsByDepth(@Nonnull final Collection<PathConfiguration> pathConfigurations) {
  return pathConfigurations.sort(false) {
    final thePath = it.path
    //'/.' means 'this node', so we don't count it as an additional directory level
    thePath = thePath[-1..-2] == "/." ? thePath - "/." : thePath
    return String.count(thePath, '/')
  }
}
```
Haven't tested..

jbornemann: When you do "Collection<PathConfiguration> modifiablePathConfig = pathConfigurations" you are just creating a new pointer to the same memory; not creating a copy, so modifiablePathConfig.sort() is still acting on the same memory of the argument. Even if you were to do a clone(), that would just clone the Collection object; The references to objects within the collection would stay the same.

Shashi-Bhushan:
Hello @jbornemann 

I've made some changes in sortPathConfigs method as you suggested with some minor modifications, and in the added Specs as well.
Please have a look at it.

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -117,6 +119,26 @@ class GrabbitConfiguration {
         )
     }
 
+    /**
+     * Removes '/.' from {@code pathConfigurations}'s path if exists in the end, and Sorts the {@code pathConfigurations}
```
jbornemann: Minor, but this line is documenting details that are apparent in the code. If I were a reader, I would be happy with just the line below, since it shows intent ("Sorts the pathConfigurations on the basis of path depth")

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -117,6 +119,26 @@ class GrabbitConfiguration {
...
+     * on the basis of paths' depth
+     *
+     * @param pathConfigurations
+     *          Collection of {@link PathConfiguration} to sort
+     * @return
+     *          Sorted Collection of {@link PathConfiguration}
+     */
+    private static Collection<PathConfiguration> getSortedPathConfigurationsByDepth(
```
jbornemann: Reason I had it as 
```groovy
List<PathConfiguration>
```
 was that List implies order, which is evident of a sorted Collection of items. 

Would need to update the JavaDoc above as well.

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -117,6 +119,26 @@ class GrabbitConfiguration {
...
+            @Nonnull final Collection<PathConfiguration> pathConfigurations) {
+
+        return pathConfigurations.sort(false) { PathConfiguration it ->
+            final String thePath = it.path
+            //'/.' means 'this node', so we don't count it as an additional directory level
+            it.path = thePath[-2..-1] == "/." ? thePath - "/." : thePath
```
jbornemann: by using, "it.path" we are modifying the original input argument's members. That is why I created "thePath" - so we can mutate it if we need to. However, I did make a mistake by making it final. Habit. 🐹 

```groovy
private static List<PathConfiguration> getSortedPathConfigurationsByDepth(@Nonnull final Collection<PathConfiguration> pathConfigurations) {
  return pathConfigurations.sort(false) { 
    String thePath = it.path
    thePath = thePath[-2..-1] == "/." ? thePath - "/." : thePath
    return thePath.count('/')
  }
}
```

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
```diff
@@ -117,6 +119,26 @@ class GrabbitConfiguration {
...
+     *
+     * @param pathConfigurations
+     *          Collection of {@link PathConfiguration} to sort
+     * @return
+     *          Sorted Collection of {@link PathConfiguration}
+     */
+    private static Collection<PathConfiguration> getSortedPathConfigurationsByDepth(
+            @Nonnull final Collection<PathConfiguration> pathConfigurations) {
+
+        return pathConfigurations.sort(false) { PathConfiguration it ->
```
jbornemann: "it" is implied in every closure if an argument name isn't provided

src/test/groovy/com/twcable/grabbit/client/GrabbitConfigurationSpec.groovy
```diff
@@ -477,4 +477,75 @@ class GrabbitConfigurationSpec extends Specification {
...
+        """
+
+        when:
+        def output = GrabbitConfiguration.create(input)
+
+        then:
+        output instanceof GrabbitConfiguration
+
+        output.pathConfigurations.eachWithIndex { element, index ->
+            assert element.path.count('/') == index + 1
```
jbornemann: Eh, I kinda like it the way you had it before. Something like this : 

```groovy
output.pathConfigurations.eachWithIndex { element, index -> 
  assert pathList[index] == element
}
```

src/main/groovy/com/twcable/grabbit/GrabbitConfiguration.groovy
Shashi-Bhushan: @jbornemann I apologize, i was under the impression that path needs to be changed in original pathConfig in case of path having '/.'. Changed it now, please have a look at the changes.

sagarsane:
@Shashi-Bhushan @jbornemann although the changes made here I don't think will break anything ... I don't think these will actually solve the issue.

Although we are sorting the pathConfigurations based on the path depth ... Since Grabbit content sync utilizes concurrent processing (currently 10 paths are kicked off in independent threads), even if we start the Jobs for the paths in the sorted order, how the threads are handled by JVM is ofcourse non-deterministic ... so there could still be cases where path `/a/b/c/d` is processed before `/a/b` ... 

It might be worth looking at the `DefaultClientService.initiateGrab()` method ...

jbornemann:
@sagarsane I think you are right. I'm sorry @Shashi-Bhushan for not recognizing this sooner. This is not going to be the full picture.

I've thought about this a bit, and I think I have a way we could solve this; However, I haven't entirely convinced myself that we should solve this problem based on how complex it may be. In case we don't solve this problem, I will also put my thoughts within the ticket for another time.

Here is how I think we can solve this problem:

1.) We need to create a ranking for paths. Here are some example paths for demonstration:

/foo                        - 1
/foo/bar                 - 2
/foo/bar/zoo          - 3
/foo/car/do            - 1

Start each path out with a ranking of 1, then for each path in sorted order based on path depth, increment the path's ranking by it's parent's ranking if it is found within the visited set; otherwise keep a ranking of 1.

This makes sure paths that can be processed independently are within the same rank group.

After that, we need to make sure Spring Batch will feed the jobs from our job queue, into the thread pool in order, based on ranking. To do this, we would need a Spring ThreadPoolTaskExecutor instance with a ThreadPoolExecutor that is built with a BlockingPriorityQueue implemented with a Comparator that compares our jobs based on path ranking. Long winded, I know..

Still even then, this would only make sure that jobs are feed into the pool in priority order, but it wouldn't prevent certain sized thread pools from picking up and starting work on higher ranked jobs before lower ranked jobs complete. To solve this, we would probably need some sort of system within our validator task that pools the job queue to see if lower ranked jobs are still in flight, and if so, put the thread to sleep for some period. 

All this to say, I think getting to a smart system for handling path dependencies would require some work. The complexity required to do so, may or may not be worth the value added.

Shashi-Bhushan:
Closing the PR, as the proposed solution is complex to implement when compared to the output result. had a discussion with @jbornemann regarding this, quoting him

> Any solution will be too complicated compared to the value it would give us. At the very minimum, what we have now (checking to see if path already has an existing parent before running the path's job - and raising a configuration error) I think is acceptable

Issue #7, opened 2015-03-10
Implement feature to Stop a Job given a jobExecution ID


jazeren1:
Possibly something like

```DELETE /grabbit/job/<JobExecutionid>```

PR #146, opened 2016-08-24
issue-7-stop-job

viveksachdeva:
Issue: https://github.com/TWCable/grabbit/issues/7

Added option to stop a job with given execution id...

Ready for review.
@jbornemann @jdigger @sagarsane @jayati-jt @Shashi-Bhushan

### DIFF
--- a/src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
+++ b/src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
@@ -22,6 +22,7 @@ import com.twcable.grabbit.GrabbitConfiguration.ConfigurationException
 import com.twcable.grabbit.NonExistentJobException
 import com.twcable.grabbit.client.services.ClientService
 import com.twcable.grabbit.resources.JobResource
+import com.twcable.grabbit.spring.batch.repository.JcrJobRepositoryFactoryBean
 import groovy.json.JsonBuilder
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
@@ -32,9 +33,17 @@ import org.apache.sling.api.SlingHttpServletRequest
 import org.apache.sling.api.SlingHttpServletResponse
 import org.apache.sling.api.servlets.SlingAllMethodsServlet
 import org.springframework.batch.core.explore.JobExplorer
+import org.springframework.batch.core.launch.JobExecutionNotRunningException
+import org.springframework.batch.core.launch.NoSuchJobExecutionException
+import org.springframework.batch.core.launch.support.SimpleJobOperator
 import org.springframework.context.ConfigurableApplicationContext
 
+import javax.annotation.Nonnull
+import javax.servlet.ServletException
+import javax.servlet.http.HttpServletResponse
+
 import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND
 import static javax.servlet.http.HttpServletResponse.SC_OK
 
 /**
@@ -49,7 +58,7 @@ import static javax.servlet.http.HttpServletResponse.SC_OK
  */
 @Slf4j
 @CompileStatic
-@SlingServlet(methods = ['GET', 'PUT'], resourceTypes = ['twcable:grabbit/job'])
+@SlingServlet(methods = ['GET', 'PUT', 'DELETE'], resourceTypes = ['twcable:grabbit/job'])
 class GrabbitJobServlet extends SlingAllMethodsServlet {
 
     //A "special" meta-jobID that allows for the status of all jobs to be queried.
@@ -145,6 +154,28 @@ class GrabbitJobServlet extends SlingAllMethodsServlet {
         response.contentType = "application/json"
         response.addHeader("GRABBIT_TRANSACTION_ID", String.valueOf(configuration.transactionID))
         response.writer.write(new JsonBuilder(jobExecutionIds).toString())
+    }
+
+    @Override
+    protected void doDelete(SlingHttpServletRequest request, SlingHttpServletResponse response)  {
+        String jobExecutionId = request.getParameter("jobId") ?: ""
+        SimpleJobOperator jobOperator = configurableApplicationContext.getBean(SimpleJobOperator)
+        if(!jobExecutionId.isLong()) {
+            log.warn "Parameter ${jobExecutionId} 'jobId' is invalid"
+            response.status = SC_BAD_REQUEST
+            response.writer.write("Parameter 'jobId' is invalid")
+            return
+        }
+        try {
+            jobOperator.stop(jobExecutionId.toLong())
+            response.status = SC_OK
+            response.writer.write("Job ${jobExecutionId} Successfully Stopped")
+        }
+        catch (Exception exception){
+            response.status = SC_NOT_FOUND
+            response.writer.write("No running job with id ${jobExecutionId}")
+        }
+
     }
 
     /**
--- a/src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
+++ b/src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
@@ -49,7 +49,7 @@ import static javax.servlet.http.HttpServletResponse.*
  */
 @Slf4j
 @CompileStatic
-@SlingServlet(methods = ['GET', 'PUT', 'POST'], resourceTypes = ['twcable:grabbit/job'])
+@SlingServlet(methods = ['GET', 'PUT', 'DELETE'], resourceTypes = ['twcable:grabbit/job'])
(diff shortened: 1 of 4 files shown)


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/146" target="_blank">ReviewNinja

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
```diff
@@ -49,7 +49,7 @@ import static javax.servlet.http.HttpServletResponse.SC_OK
  */
 @Slf4j
 @CompileStatic
-@SlingServlet(methods = ['GET', 'PUT'], resourceTypes = ['twcable:grabbit/job'])
+@SlingServlet(methods = ['GET', 'PUT', 'POST'], resourceTypes = ['twcable:grabbit/job'])
```
jbornemann: I'm kinda thinking 'DELETE' may be more appropriate in this case. What do you think @sagarsane @viveksachdeva ?

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
```diff
@@ -147,6 +147,28 @@ class GrabbitJobServlet extends SlingAllMethodsServlet {
         response.writer.write(new JsonBuilder(jobExecutionIds).toString())
     }
 
+    @Override
+    protected void doPost(SlingHttpServletRequest request, SlingHttpServletResponse response)  {
+        String jobExecutionId = request.getParameter("jobId") ?: ""
+        if(!jobExecutionId.isLong()) {
+            log.warn "Parameter ${jobExecutionId} 'jobId' is invalid"
```
jbornemann: Hmm..for the case of "all", wondering if we should provide different messaging than saying it's invalid. Really it's valid, just not supported for DELETE/POST

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
```diff
@@ -147,6 +147,28 @@ class GrabbitJobServlet extends SlingAllMethodsServlet {
...
+            response.status = SC_BAD_REQUEST
+            response.writer.write("Parameter 'jobId' is invalid")
+            return
+        }
+        try {
```
jbornemann:
```groovy
SimpleJobOperator jobOperator = configurableApplicationContext.getBean(SimpleJobOperator)
try {
  jobOperator.stop(jobExecutionId.toLong())
}
catch(NoSuchJobExecutionException ex) {
  response.status = SC_NOT_FOUND
  response.writer.write("No such job exists with id ${jobExecutionId}")
}
catch(JobExecutionNotRunningException ex) {
  response.status = SC_OK
  response.writer.write("Job already complete for id ${jobExecutionId}. Nothing to stop")
}
```

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
sagarsane: Yeah that might be better...

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
```diff
@@ -147,6 +147,28 @@ class GrabbitJobServlet extends SlingAllMethodsServlet {
...
+            SimpleJobOperator jobOperator = configurableApplicationContext.getBean(SimpleJobOperator)
+            jobOperator.stop(jobExecutionId.toLong())
```
sagarsane: we should also look to see whether any other cleanup of like the inputstream / session is needed after stopping the Job.

Take a look at `ClientBatchJobExecutionListener.afterJob()` 

Also, this will only stop the client .. server will still be wasting time and resources in sending his content over :-(

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
viveksachdeva: I used DELETE initially but changed it to POST... I thought that it is kind of an update... and I think of DELETE method as something for cleanup/removal... For me, ideal method for this would be PUT but that was used..
Open to change.. Thoughts??

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
viveksachdeva: @jbornemann : did not get this point... job id parameter needs to be sent to stop that job.. what is 'all' in this case?

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
viveksachdeva: @jbornemann : This is also something that I had earlier but changed it to single exception block... Reason I changed it:
It only looks for in memory jobs... So there could be case when job is there in the grabbit repository and we run stop for that particular id and we would get "No Such job exists".. And as stop() can only throw these 2 exceptions, I moved those to single block with one message..

jbornemann: @viveksachdeva I'm not sure what you mean. All jobs, including ones being run are backed by the grabbit repository. 

I think we should have the distinction...and at the very least do away with catching "Exception" for reasons I've mentioned before. It will just save us from having to clean this up in the future, because the Checker framework will complain about it..

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
jbornemann: Sure. Take a look at https://github.com/TWCable/grabbit/blob/master/src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy#L56

You can use "all" as a special meta-parameter to query all known jobs, e.g:
```
GET /grabbit/job/all
```

So what I'm saying here is - if someone tries to
```
DELETE /grabbit/job/all
``` 
we should not tell them that the jobID 'all' is invalid, because it's not. We should tell them deleting 'all' is unsupported or something

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
jbornemann: Yeah, I see your point.

 I initially suggested the DELETE verb for removing jobs from the repository. If you remove a job from the repository, ideally you would make sure it is stopped as well (which we aren't doing currently in the clean servlet I don't think, but..); and possibly vice-versa. So, maybe we can consolidate stopping a job, and deleting a job in the future, because really having them be seperate concepts is sort of an artificial creation..

So I guess if I had to lean one way, it might be DELETE

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
jbornemann: Oh wow, yeah, good catch.

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
viveksachdeva: ahh... that one.. ok

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
viveksachdeva: @jbornemann : Lets say I have lots of jobs in grabbit repo that were run some time in past and after that instance was restarted, then if I say stop job on one of those jobs, it will give throw NoSuchJobExecutionException instead of JobExecutionNotRunningException... 
If we decide to remove those jobs in addition to stopping, then two exception blocks would be more useful...

jbornemann: Can you please confirm that is the case, because that is contrary to the documentation : http://docs.spring.io/spring-batch/apidocs/org/springframework/batch/core/launch/JobOperator.html#stop-long-

In your scenario, the jobExecutionId would still be present within the job repository; So when Spring Batch queries the repository, it should not return NoSuchJobExecutionException..

In your scenario, I would expect JobExecutionNotRunningException..

viveksachdeva: Checked it again with older jobs.. It throws NoSuchJobException

src/main/groovy/com/twcable/grabbit/client/servlets/GrabbitJobServlet.groovy
viveksachdeva: @sagarsane : afterJob is getting executed automatically on trigger of stop()..

PR #168, opened 2017-01-09
Update to README for AEM 6.2 SP1 support

jbornemann:
FYI @viveksachdeva

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -26,7 +26,7 @@ image:https://api.bintray.com/packages/twcable/aem/Grabbit/images/download.svg[t
 * General Information
 
 ** link:{docsDir}/RELEASE_NOTES.md[Releases]
-** link:{docsDir}/AEMSupport.adoc[Dependencies & Version Support]
+** link:{docsDir}/AEMSupport.adoc[Installation & Version Support]
 
 * Using Grabbit
 
--- a/docs/AEMSupport.adoc
+++ b/docs/AEMSupport.adoc
@@ -1,7 +1,9 @@
-== Dependencies
+== What you need to run Grabbit.
 
-* AEM/CQ
-* To run Grabbit in your AEM/CQ instance, **you need to install a Fragment Bundle once per instance. It can be found link:https://bintray.com/artifact/download/twcable/aem/dependencies/Sun-Misc-Fragment-Bundle-1.0.0.zip[here]**
+* Supported AEM/CQ version per below
+* One-time installation of a system Fragment Bundle is needed. It can be found link:https://bintray.com/artifact/download/twcable/aem/dependencies/Sun-Misc-Fragment-Bundle-1.0.0.zip[here]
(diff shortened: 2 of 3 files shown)


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/168" target="_blank">ReviewNinja

viveksachdeva:
@jbornemann  : Installing this package did not fix the error for me.. Moreover it has an unmet dependency as I am not using any SP

jbornemann:
@viveksachdeva sorry, I uploaded a new one.

jbornemann:
[Grabbit-Deserialization-Firewall-Configuration-1.0.zip](https://github.com/TWCable/grabbit/files/695261/Grabbit-Deserialization-Firewall-Configuration-1.0.zip)

Any idea how to upload a generic file to https://bintray.com/twcable/aem/Grabbit#files/dependencies? 

Everything I read is about uploading packages

jbornemann:
Nevermind, got it figured out. You have to attach all files to a specific existing package version.

sagarsane:
This is interesting ... getting caught up now.

sagarsane:
Should we have the config be `-amended`? So that we don't overwrite incase the instance Grabbit getting installed to .. already has a config for the Deserialization service?

sagarsane:
I _think_ that way .. we can add only our stuff to the whitelist ... 

Also, I didn't check thoroughly but do we need to update the filter.xml on grabbit so the `/apps/grabbit/config/security` doesn't get inadvertently wiped out when Grabbit is installed / re-installed ?

viveksachdeva:
Agree with both the points from @sagarsane .. ```/apps/grabbit/config/security``` is getting wiped on on grabbit reinstall...

But on installing this package and running a sync from 6.2 -> 6.1 works fine..

jbornemann:
Thanks guys. I didn't even think about filter.xml. I've updated the package, and the PR to accommodate. 

-amended might work if we were dealing with just whitelist, but since blacklist is evaluated first, we have to remove those spring entries.

viveksachdeva:
I am able to do content sync from 6.2 -> 6.1 with this package installed but when I uninstall grabbit, this deserialization config also goes away

jbornemann:
Make sure you have the updated branch installed, and the updated package installed before you test the uninstall. Trying to uninstall when the old branch is already installed won't work.

viveksachdeva:
@jbornemann : Yes I did install the latest package with the updated branch having <exclude pattern="/apps/grabbit/extension(.*)"/>

jbornemann:
Sorry @viveksachdeva, I'm seeing what you are seeing now. Different behavior per install vs uninstall. I forgot how finicky working with the workspace filter is. I've got it down now. Promise :) Mind trying again?

viveksachdeva:
No problem @jbornemann . I will try this one tomorrow..

PR #179, opened 2017-03-01
Bug Fix + Documentation Change

jbornemann:
* Fixed case where input stream leaks
* Changed docs to refer to sending/receiving content more in terms of p2p, as it's more in line with that than client/server model. 
* Changed log names, as they no longer log just batch call sites. 

@sagarsane @viveksachdeva


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/179" target="_blank">ReviewNinja

Issue #181, opened 2017-05-01
Job failure after cleaning repository


bsaylor:
There is a problem occurring that causes jobs to fail to instantiate after partially cleaning a repository.  This happens when the repository has cleaned the low numbered (oldest) jobInstances (starting at 1), but there are still higher numbered jobInstances and the new transaction would cause jobInstances to collide with those pre-existing jobInstances.

**Exceptions**
```
PUT /grabbit/job HTTP/1.1] com.twcable.grabbit.client.services.impl.DefaultClientService Error while requesting a content sync for current Path: [/etc/xxx/yyy]
java.lang.reflect.UndeclaredThrowableException: null
org.apache.sling.api.resource.PersistenceException: Unable to create node at /var/grabbit/job/repository/jobInstances/6
```

**Steps to Reproduce**
1. Starting with a clean job repository, run a transaction with 5 paths. (you now have jobInstances 1-5).
2. Wait 60 minutes and run a second job with 5 paths. (You now have jobInstances 1-10).
3. Run the clean repository job with hours=1.  This will remove the first 5 jobInstances that are over 1 hour old and leave the second set of jobInstances that are less than 1 hour old. (You now have jobInstances 6-10).
4. Run a third transaction with 10 paths.  This transaction will fail when it tries to create jobInstance 6 which will collide with the existing jobInstances in the repository.

bsaylor:
This issue is occurring because within JcrGrabbitJobInstanceDao::getNextJobInstanceId(), the nextId is choose by getting the number of jobInstance children and adding 1.  This assumes that the jobInstance nodes start with 1 and are consecutive.  However, the cleaning function that removes old jobInstances does not reorder them.  So it is easily possible for the jobInstances to start at some higher value which then will causes a conflict.

jbornemann:
Good catch @bsaylor.

sagarsane:
Nice catch. Thanks @bsaylor !

jbornemann:
fixed by https://github.com/TWCable/grabbit/pull/182

PR #182, opened 2017-05-05
Issue #181: Set next jobInstanceID to last jobInstanceId + 1

bsaylor:
This fixes Issue #181 

It finds the highest jobInstanceId and increments by 1.  This way all jobInstances will sill be in chronological (ascending) order and later transactions will not conflict with earlier transactions after partially clearing some of the earlier jobInstances.

### DIFF
--- a/src/main/groovy/com/twcable/grabbit/spring/batch/repository/JcrGrabbitJobInstanceDao.groovy
+++ b/src/main/groovy/com/twcable/grabbit/spring/batch/repository/JcrGrabbitJobInstanceDao.groovy
@@ -257,10 +257,21 @@ class JcrGrabbitJobInstanceDao extends AbstractJcrDao implements GrabbitJobInsta
         if (!resolver) throw new IllegalArgumentException("resolver == null")
 
         final rootResource = resolver.getResource(JOB_INSTANCE_ROOT)
-        final nextId = (rootResource.children.asList().size() + 1) as Long
+
+        final lastInstance = rootResource?.children?.asList()?.max { Resource resource ->
+            final properties = resource.adaptTo(ValueMap)
+            properties[INSTANCE_ID] as Long
+        }
+        final lastInstanceProperties = lastInstance?.adaptTo(ValueMap)
+        Long nextId = lastInstanceProperties?.get(INSTANCE_ID) as Long
+        if (!nextId) {
+            nextId = 1
+        } else {
+            nextId += 1
+        }
+
         log.debug "Next JobInstance Id : $nextId"
         nextId
-
     }
 
     @Override
--- a/src/test/groovy/com/twcable/grabbit/spring/batch/repository/JcrGrabbitJobInstanceDaoSpec.groovy
+++ b/src/test/groovy/com/twcable/grabbit/spring/batch/repository/JcrGrabbitJobInstanceDaoSpec.groovy
@@ -181,4 +181,55 @@ class JcrGrabbitJobInstanceDaoSpec extends Specification {
         result.size() == 2
     }
 
+    def "CreateJobInstance for empty repository"() {
+        setup:
+        final builder =
+                node("var",
+                        node("grabbit",
+                                node("job",
+                                        node("repository",
+                                                node("jobInstances"),
+                                                node("jobExecutions")
+                                        )
+                                )
+                        )
+                )
+        final emptyMockFactory = new SimpleResourceResolverFactory(repository(builder).build())
+
+        when:
+        final jobInstanceDao = new JcrGrabbitJobInstanceDao(emptyMockFactory)
+        final result = jobInstanceDao.createJobInstance("FirstJob", new JobParameters())
+
+        then:
+        result.id == 1
+    }
+
+    def "CreateJobInstance for partial repository"() {
+        setup:
+        final builder =
+                node("var",
+                        node("grabbit",
+                                node("job",
+                                        node("repository",
+                                                node("jobInstances",
+                                                        node("2",
+                                                                property(INSTANCE_ID, 2),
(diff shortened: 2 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/182" target="_blank">ReviewNinja

src/main/groovy/com/twcable/grabbit/spring/batch/repository/JcrGrabbitJobInstanceDao.groovy
```diff
@@ -257,10 +257,21 @@ class JcrGrabbitJobInstanceDao extends AbstractJcrDao implements GrabbitJobInsta
...
+            final properties = resource.adaptTo(ValueMap)
+            properties[INSTANCE_ID] as Long
+        }
+        final lastInstanceProperties = lastInstance?.adaptTo(ValueMap)
+        Long nextId = lastInstanceProperties?.get(INSTANCE_ID) as Long
+        if (!nextId) {
+            nextId = 1
+        } else {
+            nextId += 1
+        }
```
jdigger: There appears to be a concurrency issue.

It's reading a value from the JCR and incrementing it, but it's not transactionally updating the JCR. So if another thread tried to ask for the next id at "the same time" (in however many milliseconds/seconds it took to write the update) there would be a race condition.

(Note the existing code has the same problem, just not _quite_ as pronounced.)

bsaylor: As @jdigger has pointed out, the previous code contained a concurrency issue which this update does not address.  A simple way to address the concurrency issue may be to modify the createJobInstance method (which calls this method to get a new jobInstanceId) to catch the exception that will occur when we try to create two instances with the same ID.  Then just call this function to get a new Id and retry the create.

sagarsane: Thanks @jdigger @bsaylor -- created https://github.com/TWCable/grabbit/issues/183

sagarsane:
@bsaylor -- thanks for the PR!

Can you please also add a unit test for the change you made in the `getNextJobInstanceId()` method so that we can have the bug fix documented in the tests?

src/main/groovy/com/twcable/grabbit/spring/batch/repository/JcrGrabbitJobInstanceDao.groovy
jbornemann: Does the instance ID need to be sequential at all?

sagarsane: @jbornemann if I remember correctly -- yes, because the JobExplorer's `getJobInstances()` API has a contract of returning the result in descending order from `start` to `count`

https://github.com/TWCable/grabbit/blob/master/src/main/groovy/com/twcable/grabbit/spring/batch/repository/JcrGrabbitJobInstanceDao.groovy#L170-L216

bsaylor:
@sagarsane Unit tests added.

src/main/groovy/com/twcable/grabbit/spring/batch/repository/JcrGrabbitJobInstanceDao.groovy
jdigger: The way that's implemented, contiguous sequential integers is implied. However it's external interface doesn't require it. As long as the ordering is consistent, there's no reason why that's needed.  `getJobInstance("foo", 0, 2)` and `getJobInstance("foo", 2, 2)` works for [999, 234, 452, 453] as long as it's guaranteed that the first call will not overlap with the second.

FWIW, "contiguous sequential integers" is convenient is simple situations, but maintaining that contract in concurrent distributed scenarios is... "tricky."

jbornemann: Would be nice to continue to utilize our "hashing" scheme of providing ID. If we lean on persistence to enforce order, then we might be able to dodge some of these headaches.

sagarsane: Makes sense @jdigger @jbornemann !

src/test/groovy/com/twcable/grabbit/spring/batch/repository/JcrGrabbitJobInstanceDaoSpec.groovy
```diff
@@ -181,4 +181,55 @@ class JcrGrabbitJobInstanceDaoSpec extends Specification {
...
+        final partialMockFactory = new SimpleResourceResolverFactory(repository(builder).build())
+
+        when:
+        final jobInstanceDao = new JcrGrabbitJobInstanceDao(partialMockFactory)
+        final result = jobInstanceDao.createJobInstance("FirstJob", new JobParameters())
+
+        then:
+        result.id == 3
+    }
+
```
sagarsane: 👍

sagarsane:
@bsaylor thanks for adding the tests! Can you please squash the commits?

I created #183 for some of the discussions in this PR @jdigger @jbornemann ...

bsaylor:
@sagarsane commits have been squashed.

jbornemann:
Seems to be an improvement. @sagarsane @jdigger okay with us merging?

sagarsane:
@jbornemann I'm ok with merging!

jbornemann:
merged. Thanks for taking the time to fix this @bsaylor !

PR #184, opened 2017-05-22
Upgrade Gradle to get around tls incompatibility issues when connecting to adobe nexus

jbornemann:
As seen in @bsaylor's PR. Also seeing it locally. A change in TLS configuration seems to be causing Adobe Nexus to present multiple certificates, and Gradle is too out-of-date to handle this. 

This upgrades us enough so that this issue is resolved, but no so much that our cq-package plugins become incompatible (as is the case with 3.5).

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'groovy'
 
 dependencies {
-    testCompile('org.spockframework:spock-core:1.0-groovy-2.3') {
+    testCompile('org.spockframework:spock-core:1.0-groovy-2.4') {
         exclude module: 'groovy-all'
     }
 }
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/TWCable/grabbit/pull/184" target="_blank">ReviewNinja

jbornemann:
@sagarsane @viveksachdeva

jdigger:
Regarding the plugin compatibility, see TWCable/gradle-plugin-cq-bundle#8 and TWCable/gradle-plugin-cq-package#24 and TWCable/gradle-plugin-scr#7

jbornemann:
Awesome @jdigger. I'll take a look soon. I created https://github.com/TWCable/grabbit/issues/186 to upgrade after the plugins are updated. 

I'll get this in to fix CI..

PR #191, opened 2017-07-30
Update CI to use openjdk8 and oraclejdk8 for builds

jbornemann:


sagarsane:
+1. Thanks @jbornemann !!

PR #209, opened 2018-06-14
Issue #208 Overly aggressive exclusion filter

### LINKED ISSUE #208: Overly aggressive exclusion filter
The exclusion feature which allows a node and its children to be excluded fro the sync is being overly aggressive and excluding similar sibling nodes by mistake.  See the example below for specifics.

**Example paths:**
/content/someContent/2015/1/otherContent
/content/someContent/2015/2/otherContent
/content/someContent/2015/11/otherContent
/content/someContent/2015/12/otherContent

**Example job**
```
"pathConfigurations": [
   "path": /content/someContent/2015",
   "excludePaths": [
      "1"
   ]
]
```

**Results**
This is intended to sync the content under /content/someContent/2015 and it should exclude /content/someContent/2015/1/ and its children.  However what happens is that it does exclude the "1" node, but also the "11" and "12" nodes.  This occurs when the name of an excluded node is also the prefix for a sibling node.

**What version of AEM are you running?**
AEM 6.1

**What version of Grabbit are you using?**
Grabbit 7.1.2

**How do you produce the issue?**
Perform a sync and exclude a node that would be the prefix for a sibling node.  For example exclude the node "other" and the sibling node "otherContent" will also be excluded.

**Does the issue occur consistently?** 
Yes

**Any additional details?**

bsaylor:
This fixes #208

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,20 @@
+Thanks for taking the time to fill out an issue! To help the community solve your issue, please provide answers to the following questions, along with supporting log and configuration files.
+
+
+
+**What version of AEM are you running?**
+
+
+
+**How do you produce the issue?**
+
+
+
+**Does the issue occur consistently?** 
+
+
+
+**Any additional details?**
+
+
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,11 @@ cache:
     - $HOME/.gradle
     - $HOME/protobuf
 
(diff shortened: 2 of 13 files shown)


jbornemann:
Well done. +1

@ssane

pgoodrich:
Went to fix this and then saw the pull request after I finished. :laughing: 

Thanks for doing this.  +1

sagarsane:
Thanks for merging Jeff. Thanks for fixing this @bsaylor !

PR #211, opened 2018-08-20
#2 : Updates the debug statement to only write out the node name and not the whole node

### LINKED ISSUE #2: Grabbit Server goes Out Of Memory for large content paths
Paths with large amounts of content (often seen in ```/content/dam``` ) tend run out of memory.

The problem has been observed to be alleviated by increasing the heap size (from 2GB to 4GB)

NOTE: Try out server using the raw methods on protobuf's CodedOutputStream ... rather than the node API.
Reference : MessageLite : https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/MessageLite
CodedOutputStream : https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/CodedOutputStream

sagarsane:
Closes #2 

@masroormohammed, @pgoodrich can you guys test this out when you get a chance? This is per the suggestion from @mzgubin in https://github.com/TWCable/grabbit/issues/2#issuecomment-398075823

To test this, it should be enough to try sending very large files (like large DAM uploads 50MB +) from Grabbit Server to Client.

### DIFF
--- a/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesWriter.groovy
+++ b/src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesWriter.groovy
@@ -41,7 +41,7 @@ class JcrNodesWriter implements ItemWriter<NodeProtos.Node>, ItemWriteListener {
         if (servletOutputStream == null) throw new IllegalStateException("servletOutputStream must be set.")
 
         nodeProtos.each { NodeProtos.Node node ->
-            log.debug "Sending NodeProto : ${node}"
+            log.debug "Sending NodeProto : ${node.name}"
             node.writeDelimitedTo(servletOutputStream)
         }
     }


jbornemann (approved):

mzgubin:
I also have some more changes that improve resiliency like catching some of the errors encountered to keep the jobs running. It would be great if that was perhaps put into a configuration whether to fail-fast or just log those errors as warnings and go on. I don't know enough about the best approach for something like that would be.
I also converted the servlet on the source being pulled from to handle post requests instead of get. That way the requests won't be limited by the GET URL length. This can be useful if there are a ton of exclusions as we've encountered. 
Also I made a change to how the exclusion filters are matching the paths. Currently the paths will be searched to find the exclusion string in any part of the path, instead I've made it so that the string is matched against just the startsWith instead. Not sure if that's how it's intended to be used, but perhaps that can also be configured somehow in the exclusion filter section itself.

So I have my code in a fork but I will go ahead and create a 7.1.4 branch and create a PR to be merged into there. That way you can review.

PR #214, opened 2018-08-29
Making the content syncs more resilient

mzgubin:
* Changes that improve resiliency like catching some of the errors encountered to keep the jobs running. It would be great if that was perhaps put into a configuration whether to fail-fast or just log those errors as warnings and go on. I don't know enough about the best approach for something like that would be.
* Converted the servlet on the source being pulled from to handle post requests instead of get. That way the requests won't be limited by the GET URL length. This can be useful if there are a ton of exclusions as we've encountered.
* Also I made a change to how the exclusion filters are matching the paths. Currently the paths will be searched to find the exclusion string in any part of the path, instead I've made it so that the string is matched against just the startsWith instead. Not sure if that's how it's intended to be used, but perhaps that can also be configured somehow in the exclusion filter section itself.
* Added more debug logging

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,8 +8,8 @@ group = com.twcable.grabbit
 version = 7.1.6
 
 # Please keep alphabetical
-apache_httpcore_version = 4.4.8
 apache_httpclient_version = 4.5.4
+apache_httpcore_version = 4.4.8
 cglib_nodep_version = 2.2.2
 commons_collections_version = 3.2.1
 commons_io_version = 1.4
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,6 +8,8 @@ group = com.twcable.grabbit
 version = 7.1.4-SNAPSHOT
 
 # Please keep alphabetical
+apache_httpcore_version = 4.4.8
+apache_httpclient_version = 4.5.4
 cglib_nodep_version = 2.2.2
 commons_collections_version = 3.2.1
 commons_io_version = 1.4
@@ -41,6 +43,7 @@ servlet_api_version = 2.5
 slf4j_version = 1.7.6
 sling_api_version = 2.9.0
 sling_base_version = 2.2.2
+sling_commons_json_version = 2.0.6
 sling_commons_testing_version = 2.0.12
 sling_commons_version = 2.2.0
 sling_event_version = 3.1.4
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ org.gradle.jvmargs=-XX:MaxPermSize=512m -XX:+CMSClassUnloadingEnabled -XX:+CMSPe
 bundleInstallRoot = /apps/grabbit/install
 
 group = com.twcable.grabbit
-version = 7.1.3
+version = 7.1.4-SNAPSHOT
 
 # Please keep alphabetical
 cglib_nodep_version = 2.2.2
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,8 +8,6 @@ group = com.twcable.grabbit
 version = 7.1.6
 
 # Please keep alphabetical
-apache_httpclient_version = 4.5.4
-apache_httpcore_version = 4.4.8
 cglib_nodep_version = 2.2.2
 commons_collections_version = 3.2.1
 commons_io_version = 1.4
--- a/gradle.properties
+++ b/gradle.properties
@@ -8,6 +8,8 @@ group = com.twcable.grabbit
 version = 7.1.6
 
 # Please keep alphabetical
+apache_httpcore_version = 4.4.8
+apache_httpclient_version = 4.5.4
 cglib_nodep_version = 2.2.2
 commons_collections_version = 3.2.1
 commons_io_version = 1.4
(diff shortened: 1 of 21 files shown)


gradle.properties
```diff
@@ -5,9 +5,11 @@ org.gradle.jvmargs=-XX:MaxPermSize=512m -XX:+CMSClassUnloadingEnabled -XX:+CMSPe
 bundleInstallRoot = /apps/grabbit/install
 
 group = com.twcable.grabbit
-version = 7.1.5
+version = 7.1.6
 
 # Please keep alphabetical
+apache_httpclient_version = 4.5.4
+apache_httpcore_version = 4.4.8
```
sagarsane: Since we are using the OkHttpClient already, it would be better to continue to use that than having another way to create connections using the Apache HttpClient.

gradle.properties
```diff
@@ -41,6 +43,7 @@ servlet_api_version = 2.5
 slf4j_version = 1.7.6
 sling_api_version = 2.9.0
 sling_base_version = 2.2.2
+sling_commons_json_version = 2.0.6
```
sagarsane, line 44: It would be better to continue using Groovy's built-in JSONBuilder for consistency as used elsewhere in the codebase rather than adding this new dependency!

src/main/groovy/com/twcable/grabbit/server/GrabbitContentPullServlet.groovy
```diff
@@ -35,12 +40,94 @@ import static javax.servlet.http.HttpServletResponse.SC_OK
...
+            //This user will be used to connect to JCR.
+            //If the User is null, 'anonymous' will be used to connect to JCR.
+            final serverUsername = request.remoteUser
+
+            log.info "\n\n *** About to call server service with the following to start pushing content to " +
+                    "client***\n\n" +
+                    "path={}\nafter={}\nexcludePathList={}\n\n\n", path,
+                    afterDateString, StringUtils.join(excludePathsList, ",")
+
+            serverService.getContentForRootPath(serverUsername, path, excludePathsList ?: null,
```
sagarsane, line 121: I agree the need to have a POST here instead of GET due to GET's limitations. However, since this is an internal call (from Grabbit client to Grabbit server), I don't think we need to keep both GET and POST. 

I would thus suggest removing `doGet()` and also updating the `GrabbitContentPullServletSpec` accordingly.

src/main/groovy/com/twcable/grabbit/server/services/impl/TreeTraverser.java
```diff
@@ -0,0 +1,344 @@
+package com.twcable.grabbit.server.services.impl;
```
sagarsane, line 1: Hi @mzgubin I did not see info about this in your PR description. Could you please let us know the need to re-implement this instead of using the TreeTraverser from Jackrabbit commons as it was before?

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesWriter.groovy
```diff
@@ -40,9 +41,25 @@ class JcrNodesWriter implements ItemWriter<NodeProtos.Node>, ItemWriteListener {
         ServletOutputStream servletOutputStream = theServletOutputStream()
         if (servletOutputStream == null) throw new IllegalStateException("servletOutputStream must be set.")
 
-        nodeProtos.each { NodeProtos.Node node ->
-            log.debug "Sending NodeProto : ${node.name}"
-            node.writeDelimitedTo(servletOutputStream)
+        try {
+            //log.info "\n\n### NodeProtos ###\n\n${StringUtils.join(nodeProtos.toString(), "\n")}"
```
sagarsane: We can get rid of all the commented out code (here and elsewhere in the PR). Thanks!

src/main/groovy/com/twcable/grabbit/server/GrabbitContentPullServlet.groovy
mzgubin: Ok will do

gradle.properties
mzgubin: Sure, I'll switch it

gradle.properties
mzgubin: Ok will switch to that one

src/main/groovy/com/twcable/grabbit/server/batch/steps/jcrnodes/JcrNodesWriter.groovy
mzgubin: No problem will do

mzgubin:
Sorry have been bogged down but will definitely get to this in the next week

src/main/groovy/com/twcable/grabbit/server/services/impl/TreeTraverser.java
mzgubin: Well the reason was that it allowed me to log what the error was when it was traversing. By default it was set to ErrorHandler.IGNORE so the errors were just being swallowed up.

PR #235, opened 2020-05-19
Fix for issue 153 sun.misc Import-Package

winklerj:
Changed the sun.misc package as an optional OSGi resolution.  This will remove the dependency on installing system Fragment Bundle for sun.misc and improve the installation process


winklerj:
Can I get a review on this PR?

